PR #2, opened 2017-11-12
Add TypeScript definition

kmxz:
Add `.d.ts` so it can be used in TypeScript projects.

Each individual file in `lib/` also gets its own `.d.ts` file, so that importing individual functions is also possible (similar to what https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/lodash has done).


sithmel:
thanks for you contribution! Publishing now :-)

PR #3, opened 2018-05-10
Use idiomatic JavaScript

KSXGitHub:
### Main change

Use idiomatic ECMAScript 6 syntax to make code easier for humans to read and for JS engine to optimize.

### Other chages

- Use babel-preset-env instead of babel-preset-es2015
- Create shrinkwrap.yaml (because I used pnpm to install dependencies)
- Created .npmignore
- Add 'engines' to package.json

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,7 +1,11 @@
 {
-  "plugins": ["transform-runtime"],
-  "presets": [
-    ["es2015", { "loose": true }]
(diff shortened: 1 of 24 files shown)


sithmel:
Published as iter-tools@1.5.1

KSXGitHub:
Great! thanks.

Issue #4, opened 2018-05-11
update typescript definition


sithmel:
I am in the process to add many async iterator functions, now that they are supported by V8 :champagne: 

We already have typescript definitions thanks to @kmxz . It would be nice adding the definitions to the new functions (in lib/async).

sithmel:
@kmxz consider it an help request :smile:

kmxz:
I'll be glad to help.

Just give me some time and let me know how would you like to export the async functions in `index.js`.

sithmel:
many thank, @kmxz!
I have just updated the index.js (on master)
I have also changed the arguments of "reduce", the only real breaking change.
Now it can be curried, just like map, filter etc.

PR #6, opened 2018-05-12
Add type definitions for async variants

kmxz:
Add `.d.ts` files in response to #4. 

Requires `--lib ESNext.AsyncIterable` argument to be used on TypeScript compiler.


sithmel:
published on 2.0.1. Many thanks!

PR #8, opened 2018-05-25
Rework everything to support es6 modules as build target

conartist6:
The goal of this changeset is to allow simpler and more modern inclusion patterns.

I have flattened the output so that you now do require('iter-tools/filter') in, say, node. This and other changes are almost certainly breaking so I have bumped the major version number.

It should now be possible to do import {filter, map} from 'iter-tools';
and let your packager take care of tree shaking away the rest of the lib.

I have removed npm-release, as it seems that the npm version command is meant to fulfill
all the same major functions.


.babelrc
```diff
@@ -1,11 +1,27 @@
...
-  ],
-  "presets": [
-    ["env", {
-      "loose": true
-    }]
-  ]
+  "env": {
+    "es5": {
+      "plugins": [
+        "add-module-exports",
```
conartist6: Writes module.exports = exports['default'] to the end of the transpiled files if there is only a default export.

.babelrc
```diff
@@ -1,11 +1,27 @@
...
+        "transform-async-generator-functions"
+      ],
+      "presets": [
+        ["env", {
+          "loose": true
+        }]
+      ]
+    },
+    "es6": {
+      "plugins": []
```
conartist6: Actually I think this is wrong. The library shouldn't try to be full native yet because async generators are nonstandard. It needs syntax-async-generators and transform-async-generator-functions here. Then the async generators will just get transpiled down to separate generators and asyncs (vs all the way to es5).

I guess in the past because it was encouraged not to import the whole library from index there was less of an issue with environments which don't support async generator syntax. That suggests to me that an alternative to transpiling might be to split out a separate async index which is not included in the normal index. Actually, I think that's sensible anyway.

.babelrc
```diff
@@ -1,11 +1,27 @@
...
+    },
+    "cjs": {
+      "plugins": [
+        "syntax-async-functions",
+        "syntax-async-generators",
```
conartist6: cjs also needs transform-async-generator-functions

.eslintrc.js
```diff
@@ -1,3 +1,9 @@
 module.exports = {
-    "extends": "standard"
-};
\ No newline at end of file
+  "extends": "standard",
+  "parser": "babel-eslint",
```
conartist6: Switching to the babel-eslint parser means eslint can handle the async generators correctly now.

.npmignore
```diff
@@ -1,6 +0,0 @@
-.babelrc
-.eslint*
-.gitignore
-.nvm
-.travis*
-shrinkwrap.yaml
```
conartist6: Not needed because we have shifted to opting into which files will be packaged: see npm run dist

.gitignore
```diff
@@ -1,2 +1,7 @@
-node_modules
-bower_components
+/package/*
+!/package/.gitkeep
+/node_modules
+/es5/**/*.js
```
conartist6: Why were you previously keeping your transpiler results stored in git? The only thing I could think is that it makes it easier to do development by pointing npm at this git repo. In theory doing compilation in the prepack script should make that work -- it's meant to happen after devDependencies are installed, but there's a bug out on that https://github.com/npm/npm/issues/19564. The bug has a workaround specified though, so I'll try it.

package.json
```diff
@@ -1,19 +1,19 @@
 {
   "name": "iter-tools",
-  "version": "3.0.0",
+  "version": "4.0.0",
   "description": "The iterator toolbox",
-  "main": "index.js",
-  "typings": "index.d.ts",
+  "main": "lib/index.js",
```
conartist6: This is wrong again since I shifted it to publish out of the lib dir.

package.json
```diff
@@ -1,19 +1,19 @@
...
     "benchmarks": "node benchmarks",
     "watch": "npm run test -- -w",
-    "lint": "eslint --fix --ext .js ./lib ./tests ./benchmarks",
-    "build": "babel --plugins transform-runtime lib -d es5",
-    "release:major": "npm-release major",
-    "release:minor": "npm-release minor",
-    "release:patch": "npm-release patch",
-    "precommit": "npm run build && npm run lint",
+    "lint": "eslint --fix --ext .mjs ./src ./tests ./benchmarks",
+    "clean": "rm -rf lib/*.* && rm -rf lib/async/*.* && rm -rf es5/*.* && rm -rf es5/async/*.* && rm -rf package/*",
```
conartist6: I don't destroy my directory structure, so that I don't need to have mkdirp's everywhere. Then again, mkdirp isn't all that annoying.

package.json
```diff
@@ -1,19 +1,19 @@
...
+    "build": "npm run clean && cross-env BABEL_ENV=es5 babel -x \".mjs\" src -d es5 && cross-env BABEL_ENV=cjs babel -x \".mjs\" src -d lib && cp src/*.mjs src/*.d.ts lib && cp src/async/*.mjs src/async/*.d.ts lib/async",
+    "dist": "npm run build && cp -R package.json lib/* es5 LICENSE README.md package/ && cd package && npm publish",
```
conartist6: So the new publish workflow might be:
npm version major
npm run dist

package.json
```diff
@@ -31,20 +31,23 @@
...
     "eslint-plugin-node": "^6.0.1",
-    "eslint-plugin-promise": "^3.7.0",
-    "eslint-plugin-standard": "^3.0.1",
+    "eslint-plugin-promise": "^3.8.0",
+    "eslint-plugin-standard": "^3.1.0",
     "husky": "^0.14.3",
     "measure-speed": "^1.0.0",
     "mocha": "^5.0.5",
-    "npm-release": "^1.0.0",
+    "renamer": "^0.6.1",
```
conartist6: I didn't end up using this, should drop it.

src/index.mjs
```diff
@@ -0,0 +1,51 @@
...
+export { default as regexpSplit } from './regexp-split'
+export { default as regexpSplitIter } from './regexp-split-iter'
+export { default as splitLines } from './split-lines'
+export { default as repeat } from './repeat'
+export { default as slice } from './slice'
+export { default as takeWhile } from './take-while'
+export { default as tee } from './tee'
+export { default as zipLongest } from './zip-longest'
+export { default as zip } from './zip'
+export { default as asyncIterToArray } from './async/async-iter-to-array'
```
conartist6: It might make sense to put these in ```async/index```. Quite a few people are likely to use the library only for the sync versions, so I don't think it's that big a burden to have to write:
```import asyncMap from 'iter-tools/async';```

File: src/iter.d.ts [hunk 12]
```diff
@@ -0,0 +1,2 @@
+import { iter } from './index';
+export = iter;
```
conartist6: the only change here is importing from ./index instead of ../index. It still broke rename detection though...

conartist6: Oops, I had forgotten to do this for the async typedefs too.

.eslintrc.js
sithmel: Amazing ! I didn't know that!

.npmignore
sithmel: Looking at "npm run dist" ...

conartist6:
No hurry. I'm not absolutely sure of everything I've done either.

With this last commit though it is now possible to test things out in a project just by doing ```npm install conartist6/iter-tools```. It's a bit of a kludge since the paths are different (e.g. iter-tools/lib/filter vs iter-tools/filter, and iter-tools/lib/async vs iter-tools/async), but if all you're doing with the library is named imports from the index (which I suspect will be most people) it's the same.

PR #10, opened 2018-05-29
Better transpilation

sithmel:


package.json
```diff
@@ -2,47 +2,53 @@
...
-    "benchmarks": "node benchmarks",
+    "benchmarks": "npm run build && node benchmarks",
     "watch": "npm run test -- -w",
     "lint": "eslint --fix --ext .mjs ./src ./tests ./benchmarks",
-    "clean": "rm -rf lib/*.* && rm -rf es5/*.* && rm -rf package/*",
-    "build": "npm run clean && cross-env BABEL_ENV=es6 babel -x \".mjs\" src -d lib && renamer --regex --find '\\.js$' --replace '.mjs' 'lib/**' && cross-env BABEL_ENV=es5 babel -x \".mjs\" src -d es5 && cross-env BABEL_ENV=cjs babel -x \".mjs\" src -d lib && cp src/*.d.ts lib",
-    "dist": "npm run build && cp -R package.json lib/* es5 LICENSE README.md package/ && cd package && json -I -f package.json -e \"this.main='index.js';this.typings='index.d.ts';this.module='index.mjs'\" && npm publish",
+    "clean": "rimraf dist",
+    "build": "npm run clean && npm run build:cjs && npm run build:es5 && npm run build:es5-cjs && npm run build:es2015 && npm run build:es2015-cjs && cp src/*.d.ts dist",
+    "build:cjs": "cross-env BABEL_ENV=cjs babel -x \".mjs\" src -d dist && copyfiles -f src/*.mjs dist/",
```
conartist6: I must say again that you almost certainly should not use the untranspiled files in the root of the package.

No transpilers touch code in node_modules (which iter-tools will be). However, static analysis, tree shaking, minificiation, etc, all need to be done on these files, and will fail unless the user's toolchain happens to be configured for nonstandard language features.

Finally when the untranspiled code makes it to the browser, in the event that the browser can't handle the syntax it will fail parsing /the entire bundle/. No code runs at all, and users start asking why they have just a blank white screen...

IE11 will also junk the bundle if someone packages iter-tools without reading the documentation on using the /es5 folder. The docs would have to say that you will almost certainly want to use es5 unless you definitely know what you're doing. I doubt that most people who in theory need to support IE11 actually bother testing it very often, so it could be a while before they notice that their whole app is gone.

Embrace the principle of least surprise!

package.json
```diff
@@ -2,47 +2,53 @@
...
+    "build:es5": "cross-env BABEL_ENV=es5 babel -x \".mjs\" src -d dist/es5 && renamer --regex --find '\\.js$' --replace '.mjs' 'dist/es5/**'",
+    "build:es5-cjs": "cross-env BABEL_ENV=es5-cjs babel -x \".mjs\" src -d dist/es5",
+    "build:es2015": "cross-env BABEL_ENV=es2015 babel -x \".mjs\" src -d dist/es2015 && renamer --regex --find '\\.js$' --replace '.mjs' 'dist/es2015/**'",
```
conartist6: You can get avoid having to user renamer since you've brought in babel 7. You can just use `babel --keep-file-extension` instead.

package.json
```diff
@@ -2,47 +2,53 @@
...
+    "build:es2015-cjs": "cross-env BABEL_ENV=es2015-cjs babel -x \".mjs\" src -d dist/es2015",
     "prepare": "#https://github.com/npm/npm/issues/19564",
+    "dist": "npm run build && copyfiles -f package.json LICENSE README.md dist/ && cd dist && json -I -f package.json -e \"this.main='index.js';this.typings='index.d.ts';this.module='index.mjs'\" && npm publish",
```
conartist6: So to confirm, you wish for users to write `const map = require('iter-tools/dist/es2015/map')`?

I had set up publishing from a subdirectory specifically so that at least for the common use case `const map = require('iter-tools/map')` or `import {map} from 'iter-tools'` would work.

sithmel: `const map = require('iter-tools/map')` or `import {map} from 'iter-tools'`
or using a transpiled version:
`const map = require('iter-tools/es5/map')` or `import {map} from 'iter-tools/es5'`
Did I copy something wrong?

package.json
sithmel: Would you put the es5 version in the root?

package.json
conartist6: See https://davidwells.io/blog/publishing-flat-npm-packages-for-easier-import-paths-smaller-consumer-bundle-sizes/ for what I was doing with the package directory.

sithmel: It looks to me exactly the same, except I am using the dist directory

package.json
conartist6: By my own logic, yes. I don't love it because the files are so large and gross, but I don't think the other options are better. At least it means that everyone can write their code the way they want without having to worry about IE11, and I don't hope to spend any more of my life worrying about IE11.

package.json
conartist6: Oooooh I'm so sorry, I didn't read this carefully enough the first time : (. I see now that you're still doing that.

conartist6:
Otherwise this is going to be great! I'm already using it in two projects, and I see it being a standard bit of tooling that I bring with me to many/most of the projects I work on.

PR #11, opened 2018-05-31
Fixes to README links

gabceb:


sithmel:
fixed in another branch. Thanks for spotting it!

PR #16, opened 2018-06-14
Add size, find, entries, keys, values, and change iter behavior

conartist6:
Iter used to treat objects with next as an own or proto key as iterators and returned them untouched.
This behavior is not safe, and has been removed. The only way to know if something is an iterator is to have created it yourself.


sithmel:
I like everything of this PR.
Do you also want to fix the currying here ?

conartist6:
Nah, no point mixing up conceptually separate changesets.

PR #19, opened 2018-06-14
Reorganize currying

### LINKED ISSUE #17: Currying perf
A major perf issue in this library is the implementation of currying. Currying is, in iter-tools, patterned like this:
```
export default function find (func, iterable) {
  function curriedFind (i) {
    // ...
  }
  if (iterable) {
    return curriedFind(iterable)
  }
  return curriedFind
}
```
This is problematic, because every time find is called a new function must be created.

The patterning should be:
```
export default function find (func, iterable) {
  if (!iterable) {
    return iterable => find(func, iterable)
  }
  // ...
}
```
Now in the normal situation, when all the params are passed, we don't have to create a new function.
I'm not certain of all the moving parts in play, but I suspect the current patterning might interfere with precompilation. These curriedMethod functions will get hit a lot, and if they're not all instances of the same function I don't think v8 would be able to detect them getting hot in order to do precompilation.

Also in the proposed style currying is done with an arrow function, which is lighter weight than a normal function due to optimizations of its own (doesn't create arguments, has no prototype).

conartist6:
Fixes #17


sithmel:
I tried running some tests (in the benchmark folder).
Believe it or not, this doesn't make any difference performance wise.
It still an improvement of readability and style though

conartist6:
Interesting. So for test on very large iterations (1mil items) I saw virtually not difference, I think because the by that time the function gets hot even if it was unoptimized at the beginning of the run. The key is that it's a different function each time you call map.

I designed a test case to maximize the effects of my changes by calling map repeatedly to do smaller iterations each time.
```
function mapVsBuiltinMap() {
  var a = Array.from(range(10))
  measureSpeed(function () {
    for (let i = 0; i < 1000; i++) {
      a = map(x => x + 1, a);
    }
    a = Array.from(a);
  }, { samples: 100, discard: 10 }, function (err, ms) {
    if (err) return console.log('Error!')
    console.log('************ map array ************')
    console.log(ms.toFixed(2))
  })

  var arr = Array.from(range(10))

  measureSpeed(function () {
    for (let i = 0; i < 1000; i++) {
      arr = arr.map(x => x + 1);
    }
  }, { samples: 100, discard: 10 }, function (err, ms) {
    if (err) return console.log('Error!')
    console.log('************ array.map ************')
    console.log(ms.toFixed(2))
  })
}
```

Results:
Without new changes:
************ map array ************
2.34
************ array.map ************
0.05


With new changes:
************ map array ************
0.72
************ array.map ************
0.05

My take away is that for pure speed you just can't compete with Array.map.

PR #21, opened 2018-07-01
Coverage! (and more)

### LINKED ISSUE #20: Test coverage
It would be good to get test coverage reporting set up on this library.

Test coverage is already quite high, and people shopping libraries love to see the little badge with a high coverage percentage.

conartist6:
Move from mocha to jest. This is because jest manages its own module paths, whereas mocha just relies on node's path resolution.

To test multiple code versions, we specify different path-mappers that map `require('iter-tools')` to one of the target directories.

Make some changes required when switching to Jest:
   Colocate test files in src tree (under `__test__`)
   Remove istanbul and mocha (they are built into jest)
   Replace `assert.deepEqual(a, b)` with `expect(a).toEqual(b)`.
   Replace `assert.equal(a, b)` with `expect(a).toBe(b)`
   Change test headers to `/* eslint-env node, jest */`

Rip out all the test code which manually required more than one version of a particular function! Wheee!

To get the actual coverage badges I think you'll want to use [codecov.io](https://codecov.io/gh). Should be relatively easy since Travis is already set up.

Fixes  #20


conartist6:
Looks like the tests that measure time are failing. Those are going to be flaky the way they're written now. I think I can fix them.

sithmel:
I dislike the all-in-one approach of jest. And the BDD assertion style as well (it just doesn't fit in my mind).
I have to admit though, the final result looks neat and I see the advantages. So I am ok to merge this as soon as it is fixed

COOKBOOK.md
```diff
@@ -0,0 +1,145 @@
...
+import { map, filter, compose, toArray } from 'iter-tools';
+
+const activeUserNames = compose(
+    toArray,
+    filter(name => name.length > 0),
+    map(([id, user]) => [user.first, user.last].join(' ')),
+    filter(([id, user]) => user.status === 'active'),
+    filter(Boolean),
+  )(users);
+```
```
sithmel: I would mention that, as of current js engine, for small numbers, arrays is still way faster

COOKBOOK.md
```diff
@@ -0,0 +1,145 @@
...
+```js
+const myMap = new Map(entries({
+  foo: 'foo',
+  bar: 'bar',
+}, true));
+```
+
+### Destructured iteration on Maps
+```js
+for (const [key, value] of entries(myMap, true)) {
```
sithmel:
```js
for (const [key, value] of myMap) { ...
```
would be the same ...

README.md
```diff
@@ -505,12 +425,103 @@ await asyncFind(animal => animal.kind === 'dog', [
...
+Same as tap, but for async iterables
+```js
+compose(
+  asyncTap(item => console.log(item)),
+  asyncFilter(item => !!item),
+)(asyncIter([0, 1, 2])) // logs "1", "2". returns AsyncIterable[1, 2]
+```
+
 ## size
 Returns the number of values yielded by an iterable. Without loading the whole sequence in memory. It works for both sync and async iterables.
```
sithmel: is it true that works for async-iterable? why async-size exists, then ?

sithmel:
Hi,
I am reviewing the PR. I have left a couple of comments.
I didn't see the coverage badge in the readme, are you still working on this PR?
also please add yourself as contributor in the package.json :-)

COOKBOOK.md
conartist6: Ah, almost the same, but not the same. The `true` parameter is causing entries to be reused meaning there's no allocations happening on each iteration. The same is not true of a Map's default iterator. That's the point of this bit of the cookbook!

I'll definitely add the caveat that the default behavior is fine in most situations.

README.md
conartist6: Oops, no, of course not.

conartist6:
I think I can't make the coverage badge myself because of the way Github works. I think I could if the project were part of a Github organization that I was a member of I could, but because it's your personal repository I can't give the tools the rights they need via OAuth. With this PR everything should be in readiness for you to do the last steps involving codecov.io and paste the badge image into the README.

conartist6:
Certainly, I can make more changes. I have some more locally that I'm still working on. I think the only wrong move with compose is supporting two apis. I looked at the two libraries I could think of that have compose functions: [redux.compose](https://redux.js.org/api-reference/compose) and [underscore.compose](https://underscorejs.org/#compose) and neither uses an array, but that was my only motivation for that change.

There is no real need to reset the array to 0 length, I can just keep an index variable and insert at it.

conartist6:
Just a fun tidbit, in particular because you weren't so excited about moving from `assert` to `expect`.

I am going to need to get a PR accepted to Jest first (shouldn't be a problem I hope), however then we'll have actual iterator matchers.

We can change test assertions from:
```js
expect(Array.from(iter)).toEqual([2, 4, 6])
```
into:
```js
expect(iter).toBeIterable(expect.yields(2, 4, 6)) // yields runs against an iterator taken from the Iterable
```
Also supported will be:
```js
expect(iter).yields(2, 4, 6) // Not iterable, just an iterator.
```

conartist6:
I've verified that the syntax can work and, as far as I can tell meshes with the way the library intends for things to be done.

See the proof of concept code: https://github.com/facebook/jest/issues/6649#issuecomment-403235872

sithmel:
The assertion for iterables is a great idea :+1:  (do you work in FB? out of curiosity)

The point about "reuseEntry" is still valid. Can you prove the optimisation works in practice with a tangible improvement ?
My guess is that is going to be only a marginal improvement and probably not worth it.

conartist6:
I did until recently. The reuseEntry thing isn't about the speed of the operation itself. If you're benchmarking you'd want to look for memory usage. Using less memory means more memory for everyone else, and less work down the line for the garbage collector.

In my application, a game, I have a board. The board has 600 squares. Board logic executes every 100ms, and each square has 4 layers. I need to pass the board from the game logic layer to the rendering layer, which I use iterators to do. My logic to do this is essentially that of `zip` in order to pass the data between application layers in a structured fashion, however I implement zip and the reuseEntries optimization by hand. If I didn't I'd be allocating 6000 length-4 arrays every second.

I think iterators as a pattern are most powerful when working in tight loops, and giving programmers a performant way to structure data in such loops seems like a win.

sithmel:
I understand your point of avoiding temporary arrays to be garbage collected. But I would like some benchmark before introducing this option in the api. Garbage collection is supposed to be very fast, and js engines performs various optimisations.
I tried to run a few variations:
```js
// average: 3180
function * zip (iterables) {
  const iters = iterables.map(i => iter(i))
  let zipped = []
  while (true) {
    zipped.length = 0

    for (const iter of iters) {
      const {done, value} = iter.next()
      if (done) return
      zipped.push(value)
    }
    yield zipped
  }
}

// average: 2060
function * zip (iterables) {
  const iters = iterables.map(i => iter(i))
  let zipped = []
  while (true) {
    let i = 0
    for (const iter of iters) {
      const {done, value} = iter.next()
      if (done) return
      zipped[i++] = value
    }
    yield zipped
  }
}

// average: 1980
function * zip (iterables) {
  const iters = iterables.map(i => iter(i))
  let zipped = new Array(iterables.length)
  while (true) {
    let i = 0
    for (const iter of iters) {
      const {done, value} = iter.next()
      if (done) return
      zipped[i++] = value
    }
    yield zipped
  }
}

// average: 2120
function * zip (iterables) {
  const iters = iterables.map(i => iter(i))
  while (true) {
    let zipped = []
    for (const iter of iters) {
      const {done, value} = iter.next()
      if (done) return
      zipped.push(value)
    }
    yield zipped
  }
}

// average: 2060
function * zip (iterables) {
  const iters = iterables.map(i => iter(i))
  while (true) {
    let zipped = new Array(iterables.length)
    let i = 0
    for (const iter of iters) {
      const {done, value} = iter.next()
      if (done) return
      zipped[i++] = value
    }
    yield zipped
  }
}
```
Sadly I realised perf_hooks (https://nodejs.org/api/perf_hooks.html) do not allow to measure garbage collection as of now. 

I would hold off introducing this feature until we are able to prove how much faster is going to be.

PR #26, opened 2018-07-08
simplify api

sithmel:
I have pretty much removed the reuseEntry optimisation and changed the "chain" signature.
The main rationale is, keeping the api simple.
Happy to re-add the optimisation when we are able to measure the advantages of it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,7 +126,7 @@ Takes in a plain object, null, a Map, or any other object which defines an `entr
 When given an Object, it is equivalent to Object.entries, otherwise it calls `entries()`
 When passed a nullish value, returns an empty iterable.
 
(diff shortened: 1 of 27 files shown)


README.md
```diff
@@ -126,7 +126,7 @@ Takes in a plain object, null, a Map, or any other object which defines an `entr
 When given an Object, it is equivalent to Object.entries, otherwise it calls `entries()`
 When passed a nullish value, returns an empty iterable.
 
-Entries takes an optional second parameter, `reuseEntry`, which defaults to false. If enabled, `entries` will emit the same entry item on each iteration step, changing that entry's values each time. If the target object is Map-like, `entries(target, true)` will attempt to call `target.keys` and `target.values` instead of `target.entries`. _If you are not certain this is safe to do, do not do it!_. Look at [Cookbook#reusing-entries](https://github.com/sithmel/iter-tools/blob/master/COOKBOOK.md#reusing-entries) for more information.
+If the target object is Map-like, `entries(target, true)` will attempt to call `target.keys` and `target.values` instead of `target.entries`. _If you are not certain this is safe to do, do not do it!_.
```
conartist6: Yeah, this shouldn't still exist. There's no point if we're not reusing entries.

Also that whole section of the cookbook should be removed too.

sithmel:
@conartist6 do you have other outstanding changes? or I would publish this as a new major version

conartist6:
This is great! Can you give me a little more time to look and think? It'll be better than releasing more major versions.

PR #29, opened 2018-07-20
Make iter and asyncIter private, add iterable and asyncIterable

### LINKED ISSUE #23: Consider deprecating iter
I think `iter` should probably not be part of the library's public API surface.

Strings and arrays are already iterables. A deprecation notice could warn users when an object is passed to `iter`, informing them that they should use entries instead.

Internally a function called iter could continue to be used with the same basic responsibility of making an iterator out of an iterable.

conartist6:
Fixes #23


conartist6:
Actually, I originally said deprecate iter and asyncIter, but this diff removes them. There may be enough people using those functions to warrant keeping them around with a deprecation warning.

sithmel:
Using semver and a changelog looks gentle enough. Do you think a deprecation warning is necessary ?

conartist6:
Probably not necessary, but it seems like the polite thing to do. I guess usually the deprecation would go out in a non-major release. It seems like a convenient way of showing people that the new methods are different from the old ones.

conartist6:
Sorry I keep tacking on commits. I'm done now.

sithmel:
ok, I am going to review and merge ... as soon as I have 10 minutes

src/internal/iterable.mjs
```diff
@@ -0,0 +1,12 @@
+const emptyArr = []
+export default function iterable (i, ...args) {
```
sithmel: we can remove args here

src/internal/async-iterable.mjs
```diff
@@ -0,0 +1,9 @@
+import iterable from './iterable'
+
+export default async function * asyncIterable (i, ...args) {
+  if (i && i[Symbol.asyncIterator]) {
```
sithmel: we can remove args here

conartist6:
Renamed. Removed unnecessary ...args. Fixed a bug in asyncZip.

PR #41, opened 2018-07-29
added first/rest

sithmel:
**Work in progress**

This is inspired to car cdr in lisp (https://en.wikipedia.org/wiki/CAR_and_CDR) and first/rest in clojure (https://clojuredocs.org/clojure.core/first).


src/first.mjs
```diff
@@ -0,0 +1,7 @@
+import ensureIterable from './internal/ensure-iterable'
+
+export default function first (iterable) {
+  const iter = ensureIterable(iterable)[Symbol.iterator]()
+  const firstItem = iter.next()
```
conartist6: This gets a bit interesting. If the only usage of an iterable is taking the first item the underlying iterator needs to be notified, otherwise a leak occurs. Generators work this way -- they stay in a suspended execution state and AFAIK can't be garbage collected until they are finished yielding all their values. See: http://exploringjs.com/es6/ch_iteration.html#_closing-iterators-via-return

Come to think of it I think `zip` has this problem too, although not `zipLongest`.

You can easily do something like `if(typeof iter.return === 'function') { iter.return(); }`, but this would certainly be a break with `car`/`cdr` since using `first` would prevent you from then using `rest`.

conartist6:
I definitely like the methods though. They provide added readability over slice, when the selecting or omitting the first item. If you implement them like the other methods, using generators for/of and break, you won't have to worry about return (or catch) either.

src/first.mjs
sithmel: Completely unaware of this.
It probably impact many functions, I'll have a look into it

src/rest.mjs
```diff
@@ -0,0 +1,7 @@
+import ensureIterable from './internal/ensure-iterable'
+
+export default function rest (iterable) {
+  const iter = ensureIterable(iterable)[Symbol.iterator]()
+  iter.next()
+  return iter
```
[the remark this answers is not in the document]
sithmel: True, I could implement it as "slice(1)". I am going to think about it

src/first.mjs
conartist6: I'm pretty certain it doesn't. I've audited the codebase for it before. I missed zipLongest when I did, but otherwise it's good. The case only affects manual iteration (calling `next()`) when the source iterable is not exhausted.

sithmel:
I got rid of "rest". But I kept "first". I also mitigated some of the issues related to https://github.com/sithmel/iter-tools/issues/42
and created failing testcases for them.
I introduced a watch task, it is helping me a lot with the tests.

sithmel:
Still work in progress on fixing #42 bit I found the right way I think (already applied to zip, zipLongest)

src/__test__/zip.test.js
```diff
@@ -16,6 +18,36 @@ describe('zip', function () {
...
+    const iter = zip(range(2), {[Symbol.iterator]: () => repeatX})
+    expect(Array.from(iter)).toEqual([[0, 'x'], [1, 'x']])
+    expect(calledClose).toBe(true)
+  })
+
+  it.skip('closes when stopping earlier, using slice', function () { // broken in es5 version
+    let calledClose = false
+    const repeatX = {
+      next: () => ({ done: false, value: 'x' }),
+      return: () => {
```
conartist6: You can do:
```
const repeatX = {
  return: jest.fn(() => ({ done: true }),
  [Symbol.iterator]() { return this }
}
// ...
expect(repeatX.return.toHaveBeenCalled())
```

src/__test__/tee.test.js
```diff
@@ -21,6 +21,19 @@ describe('tee', function () {
     expect(iters[2].next().value).toBe(2)
     expect(iters[2].next().done).toBe(true)
   })
+
+  it.skip('closes when exhausted', function () {
```
conartist6:
```js
export default function tee (iterable, number) {
  number = number || 2
  const sourceIter = ensureIterable(iterable)[Symbol.iterator]()
  const arrays = Array.from(map(() => new Dequeue(), range(number)))
  let done = false
  let exhausted = 0

  function fetch () {
    const newItem = sourceIter.next()
    if (newItem.done) {
      done = true
    } else {
      arrays.forEach((ar) => ar.push(newItem.value))
    }
  }

  function maybeCloseIterator() {
    if (exhausted === number && !done && typeof sourceIter.return === 'function') {
      sourceIter.return()
    }
  }

  function * teeGen (a) {
    try {
      while (true) {
        if (a.length) {
          yield a.shift()
        } else if (done) {
          return
        } else {
          fetch()
        }
      }
    } finally {
      exhausted++
      maybeCloseIterator();
    }
  }
  return arrays.map((a) => teeGen(a))
}
```

sithmel:
I think this is probably good enough to merge now

PR #48, opened 2018-08-09
added negative indecies to slice

sithmel:


conartist6:
Looks good now.

conartist6:
Oh, and be sure to run coverage before you commit. Might want to add that to a hook of some kind.

sithmel:
I added a circular buffer implementation. I didn't find one I like on npm, so I wrote a simple one (only 14 LOC).
I think we can do better when "start" and "end" are both negative. I am having a look at it

conartist6:
Might want to make your buffer iterable... I guess in that case you could if you made a buffer of end-start size.

PR #50, opened 2018-08-23
adding async-buffer

sithmel:


conartist6:
What is it for... ? Your example seems to me to speak volumes, since it doesn't appear to do anything. Not that the idea isn't interesting and potentially useful, I just think you need to actually expose the buffer.

Maybe something like:
```
const {buffer, iterable} = await asyncBuffer(3, asyncIterable(range(10)));
console.log(buffer); // Iterable[1, 2, 3]
iterable.next();
console.log(buffer); // Iterable[2, 3, 4]
```

sithmel:
sorry, tests don't mean anything and implementation is wrong too.
The intention was to wrap an async iterable in another that consume the first buffering a certain amount of items in memory.
For example:
```js
const buffered = asyncBuffer(3, iterable)
const buffer = buffered()
buffer.next() // yield the first item from "iterable" and starts retrieving the 2nd and 3rd
buffer.next() // yield the second item from iterable and starts retrieving the 4th
buffer.next() // yield the third item from iterable and starts retrieving the 5th
...
```

conartist6:
OK, but it still doesn't make sense. Your example suggests that Buffer returns a sync iterable. It can't ever work, even just to get the first item, because it will need to come up with an async item synchronously.

If in fact that was a typo and the resulting iterable is meant to be asnyc, I still don't see the point. Up until the last b items (b being buffer size) it will still have to wait every time, it will just be waiting for the `i+b`'th item instead of the `i`'th. And even those last `b` items are stuck behind an async API which precludes their synchronous consumption.

conartist6:
You can't avoid the fact that a streaming buffer has both sync (access to the buffer) and async (knowing when the buffer is sufficiently full) aspects to its behavior.

sithmel:
The purpose is not to expose the buffer directly. But "to buffer" the input iterable.
This only make sense with async iterables.
Using the youtube metaphor: the resulting iterable is still returning the same video, but it pre-fetches a certain number of seconds to keep the reproduction smooth.

conartist6:
But it doesn't keep the reproduction smooth! In your analogy it prefetches 10 seconds of video and then then network is lost for five seconds. It's not going to dip into 5 seconds worth of its buffer, it's just going to take 10 seconds to start playing the video at all and it's still going to wait 5 seconds in the middle. This is because it is always forced to wait until the buffer is full to yield an item in the buffer.

Just look at the code:
```
      const { done, value } = await buffer.shift() // always waits ...
      if (done) return
      yield value // ... to yield
```

sithmel:
Let's assume the buffer size is 2.
This is what should happen:
* the for loop pre-fetch promise 1 and 2 in the buffer
* entering in the while loop
* pre-fetch promise 3 in the buffer
* extract promise 1, wait to be fulfilled and yield
* pre-fetch promise 4 in the buffer
* extract promise 2, wait to be fulfilled and yield
* ... and so on

When I call iterator.next() I get a promise and asynchronous code is scheduled to be executed.
After yielding from this async iterable, this asynchronous code has a chance to be executed.

conartist6:
Oooooooooh. I misread the code. Sorry. Particularly I didn't notice that since this code isn't doing `for .. await` it is stacking up Promises.

After reviewing the async iterator spec, I see now that the spec does not explicitly disallow (or allow) async generator functions which do not work correctly when the next record is requested before the previous result has fulfilled. I think it is, however, expected that some implementations (by hand) of the AsyncIterator interface by hand, will not work correctly in that case.

Async generator functions though are defined to, essentially, treat next calls as requests [and keep an internal buffer of them](http://tc39.github.io/proposal-async-iteration/#table-internal-slots-of-asyncgenerator-instances), which behavior it seems like this code essentially relies on.

Thus I propose a rather simple change: instead of using `internal/ensure-async-iterable`, which *may not wrap the given iterable in an async generator*, just always wrap the given iterable in an async generator, along the lines of:
```
async function * wrap(iterable) { yield* iterable }
```

Sorry it took me so long to see what was going on, I've never read that part of the spec.

sithmel:
no worries. The change you propose makes sense.

conartist6:
Just a thought, but those flaky tests could probably be rewritten using [jest timer mocking](https://jestjs.io/docs/en/timer-mocks.html)

sithmel:
I gave up after a few attempts with fake timers.
The code is a combination of timers and promises, this is not making testing easy.

PR #53, opened 2018-09-11
Redirect native modules to es2018

KSXGitHub:
Every Node.js version that supports '*.mjs' also fully supports es2018, so why not switch to es2018 to reduce bundle size and gain V8's performance benefit?

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,8 +3,8 @@
   "version": "6.1.1",
   "description": "The iterable toolbox",
   "main": "es5/index",
-  "module": "es5/index.mjs",
-  "jsnext:main": "es5/index.mjs",
+  "module": "es2018/index.mjs",
+  "jsnext:main": "es2018/index.mjs",
   "typings": "src/index.d.ts",
   "scripts": {
     "lint": "eslint --fix --ext .mjs,.js ./src ./src/__test__ ./benchmarks",


conartist6:
The reason not to do it is the web. There's a lot of stuff in this library, so I think the "pay only what you need" principle in terms of size is pretty important for the web. Create-react-app, for example, will load .mjs files before .js files, and as it uses webpack it will tree shake away unused code in production. It can only succeed in pruning the unused parts of iter-tools if it consumes the mjs files which use es6 import export syntax. create-react-app also expects that it will not need to transpile node modules in order for them to be safe for use, which means es5.

I'm extra hesitant about making es2018 a default since async iteration is still a draft proposal https://github.com/tc39/proposal-async-iteration. At the moment I would consider any usage of the es2018 sources to be experimental.

KSXGitHub:
I didn't know of this weird behavior of create-react-app.

What's about `"module:esnext"`? That's safe to keep right?

conartist6:
I like adding it, but could it point to es2015? Maybe wait to see what @sithmel says before you make that change though.

Browsers won't actually ever see the import sytnax. Almost all code shipping to browsers these days is run through bundlers like webpack or rollup which create a single file that does not have any import or export keywords in it. This way web developers are writing code in modules but still supporting Internet Explorer.

sithmel:
both "module" and "jsnext:main" are ignored by node. Node uses only "main".

"module" and "jsnext:main" are used by webpack, rollup etc.. 
By default they don't transpile dependencies, for this reason it is better using es5 by default.

PR #54, opened 2018-09-11
[TypeScript] Fix missing type definition in TypeScript consumers

KSXGitHub:
### Changes

  - Create script `"build:copyts"` which copies all *.ts files into production folders
  - Add `"build:copyts"` to `build`
  - Redirect `"typings"` to `"es5/index.d.ts"` (type definition should parallel its JS counterpart)

---

I was trying to use iter-tools in my TypeScript project but the compiler couldn't find type definition.


sithmel:
thanks a lot. Happy to merge

PR #55, opened 2018-09-11
[TypeScript] Fix and improve type definition

KSXGitHub:
### Fix broken type definition

I find some type such as `combinations` and `permutations` return `Iterable<T>` (array of scalars) instead of `Iterable<T[]>` (array of arrays). This makes no sense to me, so I decided to fix it.

### Improve type definition

TypeScript 3.0+ allows users to use tuple generic arguments which I take advantage of to have more accurate function annotations.

Function definitions I changed are: `combinations`, `combinationWithReplacement`, `permutations`, and `product`.

### Added dependencies

I added [typescript-tuple](https://github.com/ksxnodemodules/typescript-tuple) as a production dependency (which I created and maintain) to utilize its generic utilities (such as `Repeat`, `Prepend`, `Reverse`).

### Potential Breaking Changes

JavaScript users are unaffected.

On TypeScript side, this change requires TypeScript version 3.0 or later, but it's OK since TypeScript is completely useless prior to https://github.com/sithmel/iter-tools/pull/54 ~(i.e. right now)~.

**TL;DR** No breaking changes.

### Additional Changes

This pull request includes changes from https://github.com/sithmel/iter-tools/pull/54 because any change of type definition without type definition being detected is useless.

### Additional Requests

1. Is it OK if I add myself as a CODEOWNER of `*.ts` files? Because I want to make sure no one give a broken type definition.

2. If this were to be merged, I ask you to merge to a "experiment" branch and publish it under "rc" tag (so that I can install it) in npm so that I can do more testing in order for it to work.

PR #61, opened 2018-09-14
Test TypeScript definition

KSXGitHub:
This pull request provides a way to test `src/index.d.ts`.

### Why?

Prior to https://github.com/sithmel/iter-tools/pull/55, I find TypeScript definitions of this library is very inaccurate, so I feel the need to fix it. I also want to take advantage of TS 3.0 to improve the types, but I did it bindly, hoping it would work. Now with this change, everyone can add tests to verify if TS definition is correct.

### Changes

* Install [`static-type-assert`](https://github.com/ksxnodemodules/static-type-assert) (author: me) as a dev dependency
* Update `typescript` dependency to `3.0.3`
* Create script `"test:types"` and add it to `"test"`

### How would it works?

1. Tester writes type assertion statements into `src/__spec__/types.spec.ts`.
2. Run `tsc`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,5 +6,5 @@
 
 /coverage/*
 !/coverage/badge-*.svg
-__spec__/*.js
-__spec__/*.d.ts
+/src/__spec__/*.js
+/src/__spec__/*.d.ts
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,5 @@
 
 /coverage/*
 !/coverage/badge-*.svg
+__spec__/*.js
+__spec__/*.d.ts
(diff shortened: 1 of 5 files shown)


sithmel:
I found this very interesting, but I am no position to make any technical consideration on this. I am not a typescript dev.

I would go on merging on trust.

Also @dmorgantini you might be interested in this.

@conartist6 do you have any opinion on this ?

KSXGitHub:
@sithmel You're still there?

KSXGitHub:
Take your time.

sithmel:
Included in #64

PR #62, opened 2018-09-14
[TypeScript] Impose limit on parameter r of permutations/combinations functions

KSXGitHub:
### Why?

I had created a change in `index.d.ts` to improve various functions such as `combinations`, basically, it will return `Iterable<[T, T, ...]>` instead of ambiguous `Iterable<T[]>`.

But it comes with a cost: The length of each tuple in the returning iterable depends on `R` (generic type parameter of `r`), it's works fine when `R` is `number` or an integer (or a union of integer) within the range of 0 - 99, but when it isn't, it causes TSC to run into infinite loop.

Therefore, I made this change to make TypeScript treats `R` differently based on its value.

### How would it works?

I created two types: `ReasonableNumber` and `CombinationsPermutations`; `ReasonableNumber` is a union of `0 | 1 | ... | 31`, and `CombinationsPermutations` is type of `combinations`/`permutations` functions.

```typescript
type ReasonableNumber = UnionRange<32>;

type CombinationsPermutations = <T, R extends number>(iterable: IterableLike<T>, r: R) =>
  Iterable<R extends ReasonableNumber ? Repeat<T, R> : number[]>;

declare const combinations: CombinationsPermutations
...
```

When user call these function (like so: `combinations(iterable, r)`), the returning value will have type of:
  * `[T, T, ..., T]` (tuple) when `r: R` is within 0 - 31, the number of elements within tuple depends on value of `r`.
  * `T[]` (unbounded array) when `r: R` is not within 0 - 31 (e.g. `-1`, `2.5`, `1000`, `number`).
  * `[T, T, ..., T] | [T, ...] | ...` (union of tuples) when `R` is a union of number constant within 0 - 31.

### Additional Changes

* This pull request includes https://github.com/sithmel/iter-tools/pull/61 because I don't want to code bindly without tests.

### Additional Requests

* If this were to be merged, it should be published under `rc` tag first. I want to double-check it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,5 +6,5 @@
 
 /coverage/*
 !/coverage/badge-*.svg
-__spec__/*.js
-__spec__/*.d.ts
+/src/__spec__/*.js
+/src/__spec__/*.d.ts
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,5 @@
 
 /coverage/*
 !/coverage/badge-*.svg
+__spec__/*.js
+__spec__/*.d.ts
--- a/package-lock.json
+++ b/package-lock.json
@@ -7026,6 +7026,15 @@
         }
       }
     },
+    "static-type-assert": {
+      "version": "2.0.0",
+      "resolved": "https://registry.npmjs.org/static-type-assert/-/static-type-assert-2.0.0.tgz",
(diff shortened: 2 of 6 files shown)


sithmel:
Included in #64

Issue #63, opened 2018-09-14
[TypeScript] The functions should return `(Async)IterableIterator` instead of `(Async)Iterable`


KSXGitHub:
When read section "Issues and Limitations" section of README.md, I see that the functions return an iterator object (`IterableIterator` in TypeScript) instead of an interable, although iterator object is also iterable, it lacks `[Symbol.iterator]()` method and therefore does not match `Iterable` interface.

I could've make a pull request right now but I don't want to have a merge conflict with my last pull request so I create this issue as a reminder.

PR #66, opened 2018-09-16
[TypeScript] Fix groupBy/asyncGroupBy and returning types of other functions

KSXGitHub:
* Fix https://github.com/sithmel/iter-tools/issues/63
* Add missing overloads of `groupBy`/`asyncGroupBy` in which they takes `null` as the first parameter
* Fix returning type of `asyncGroupBy`: it should be `AsyncIterableIterator`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,5 +6,5 @@
 
 /coverage/*
 !/coverage/badge-*.svg
-__spec__/*.js
-__spec__/*.d.ts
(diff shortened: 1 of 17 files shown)


sithmel:
iter-tools@6.1.4-rc.2

conartist6:
So glad we got this work done, thanks @KSXGitHub!

PR #67, opened 2018-09-18
[TypeScript] Fix some types and update tests

KSXGitHub:
* **Fix some types**: I made a mistake in which `repeat` returns an array. I also forgot to fix return types of `combinations`/`permutations`.

* **Update tests**: In the [last pull request](https://github.com/sithmel/iter-tools/pull/66), I made functions return `IteratorIterable` but forgot to update corresponding tests.


sithmel:
iter-tools@6.1.4-rc.3

PR #68, opened 2018-09-18
Improve keys/values/entries; remove IterableLike

conartist6:
We use the `IterableLike` type thoughout our defs, but we actually removed support for everything but iterables from pretty much all iter-tools methods.

`IterableLike` is defined as: `type IterableLike<T> = Iterable<T> | T[] | { [key: string]: T; } | { [key: number]: T; };`

Actually T[] is a subset of Iterable<T> though, and both the object dictionary types would be runtime errors.

### DIFF
--- a/src/__spec__/entries.spec.ts
+++ b/src/__spec__/entries.spec.ts
@@ -1,0 +1,14 @@
+import assert from "static-type-assert";
+import * as iter from "../index";
+
+assert<
+  IterableIterator<[string, number]>
+>(iter.entries({foo: 42}));
+
+assert<
+  IterableIterator<[string, string | null]>
+>(iter.entries({foo: "", bar: null}));
+
+assert<
+  IterableIterator<[string, never]>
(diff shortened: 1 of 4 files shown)


sithmel:
@KSXGitHub you asked to be aware of any changes in type definitions. Here you go :smile: 

@conartist6 I'll try to understand this during next weekend

KSXGitHub:
LGTM

sithmel:
6.1.4-rc.4
@KSXGitHub please let me know when we are ready for releasing officially

KSXGitHub:
> please let me know when we are ready for releasing officially

I can't really know when it is ready. Other projects usually make their `rc` versions official after a certain amount of time (for example: release latest bug-free `rc` every 2 days). That's why it is called a schedule.

sithmel:
so I guess the current version has no issues at the moment.
I am going to release it next week, so I'll have time to catch up with all these changes

PR #71, opened 2018-09-23
Use tslint-config-standard instead of tslint:recommended

KSXGitHub:
## Pros

* To be consistent with JavaScript code: This project uses [`standard`](https://standardjs.com/) for its JS code and [`tslint-config-standard`](https://github.com/blakeembrey/tslint-config-standard) mimics standard's style.

* Some of `tslint:recommended` rules make no sense (such as `trailing-comma`).

## Cons

* Other contributors might have to change their coding habits.

---

What do you think, @conartist6?

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7354,6 +7354,65 @@
         "tsutils": "^2.27.2"
       }
     },
+    "tslint-config-standard": {
+      "version": "8.0.1",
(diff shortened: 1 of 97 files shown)


sithmel:
tslint was a clumsy attempt from my part to figure out ts issues. Using same conventions used in the code is a good idea

PR #78, opened 2018-12-01
When

conartist6:


sithmel:
I can see the value of the pattern but there are a few things that stops me from merging this.
The first one is the syntax: it is quite different from the rest of the library. And doing that breaks the principle of least astonishment.

The function does something that is achievable with a ternary expression. I know ternary expressions are ugly, but they are immediately understandable by any js developer. Using this pattern requires the developer to look at the readme instead.

Another things. The pattern makes sense with "concat", the others just are not about iterables.

Maybe this could find a space in a standalone library ... or be applied here with a different api.

For example:
```js
when(condition, iterable)
```
when "condition" is true it returns an ithe terable when false "[]".

conartist6:
Yeah that API is the simplest and was the first one I thought of, but I dismissed it because while the function declaration looks fine, the usage will be too cryptic. In particular there needs to be a verb somewhere to answer the question, "When condition: do what?" Three could be three separate functions called `concatWhen`, `yieldWhen`, and `returnWhen` I suppose.

conartist6:
Didn't meant to click that button. I also agree that ternaries are theoretically more immediately evident to new devs, and they may still be preferable in certain situations, but particular as more code is being run through prettier, that pattern gets larger and heavier. Frequently the several pieces of the expression will end up with their own largely useless lines. Once it's split across several lines, reading it involves a bit of jumping around: first you see the condition, then you probably need to skip to see the bottom of the expression which returns your iterable to see the `: []` bit, then you can jump back up to the top to see what gets emitted on the main branch. I think the cost incurred by forcing the user to go to the readme if they're really unsure of what's going on is worthwhile, but I think the fact that the result is still fed into the spread operator gives them a huge clue, and the idea of the literate syntax was to make it so obvious what was going on that even first time readers would likely just understand.

As a final note, there's actually no incompatibility between the `when(condition, iterable)`/`when(condition)(iterable)` API and the API proposed here. It would be easy enough for me to make the version proposed into this diff for simple sugar on top of the API that other methods in the library use.

conartist6:
Yeah I mean I could publish it under some other name.

Issue #75, opened 2018-11-25
Adding here a few ideas for new features


sithmel:
## Combinatory generators length
It would be nice to have handy functions to calculate the length of a combinatory generator (given the length of the input sequence). This way I can pick the nth item (for example).

## flatten
A simple flatten (only one level), can be useful. For example I could use ```flatten(zipAll(...))``` to interleave many iterators. You can do the same with "chain" and the spread operator, but would require to expand the argument in an array and it would be inefficient

## window
```js
window(3, range(100))
[0, 1, 2] [1, 2, 3] [2, 3, 4] [3, 4, 5] etc.
```
Better to pick up a different name though

## collate
It takes as arguments n sorted iterable. It returns a single sorted iterable.

## Reduce variant
Similar to the reduce but returning a iterable that return the accumulator.
I don't know yet how to call it.

conartist6:
I want a really simple one, perhaps called `when`, which would take a condition and an item, and would emit an iterable of 0 or 1 items depending on the value of condition. It would only be useful to the extent that it was more compact and readable than using a normal ternary.

For example:
```js
[
  ...(condition1 ? [item1] : []),
  ...(condition2 ? [item2] : []),
  // vs.
  ...when(condition3, item3),
  ...when(condition4, item4),
]
```

KSXGitHub:
> A simple flatten (only one level), can be useful.

Why only one level though? ES native's `array.flatten(n)` accept any level.

I suggest this:

```javascript
flatten(iterable) // flatten infinitely (until meeting a non-iterable)
flatten(n, iterable) // flatten n levels
flatten(fn, iterable) // flatten until fn(x) returns false (x is element/sub-element of iterable)
```

sithmel:
This looks a reasonable implementation though: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat

sithmel:
Added #79 #80

KSXGitHub:
I may have to correct your TypeScript declarations.

conartist6:
There's nothing conceptually wrong with not treating strings as iterables for the purpose of a flatten, even though they technically are. I'd argue that even a not arbitrarily deep flatten implementation would want to do that. Obviously if you did let flatten into a string it would blow up as the string iterator returns other strings, which becomes an infinite recursion.

[...]

PR #79, opened 2018-12-05
added flat function

sithmel:


KSXGitHub:
As you said in https://github.com/sithmel/iter-tools/issues/75#issuecomment-444634511, flatten infinitely seems impossible, but what about `flatten(fn, iterable)`?

src/flat.mjs
```diff
@@ -0,0 +1,29 @@
+import ensureIterable from './internal/ensure-iterable'
+
+function * flat (depth, iterable) {
+  if (depth === 0) {
+    yield * ensureIterable(iterable)
+  } else {
+    for (const iter of ensureIterable(iterable)) {
+      if (iter[Symbol.iterator]) {
```
KSXGitHub: Should it be `if (typeof iter[Symbol.iterator] === 'function')`?

sithmel: This is based on the assumption that if you implement ```iter[Symbol.iterator]``` with anything different from a function or undefined, you are a bad person and deserve to spend an afternoon debugging. But I am going to fix it for sake of readability :smile:

conartist6: The whole check looks superfluous to me, it's already done in `ensureIterable` so we will always end up in the first branch here.

sithmel: I am not sure I follow.
ensureIterable ensures that the original iterable is an iterable. Then I check item by item if they are iterable or not.

conartist6: I'm sorry, I was reading wrong. The code is right.

sithmel: Better safe than sorry. Thanks for checking ;-)

sithmel:
@KSXGitHub can you please have a look at the TS signature? you can now use a function instead of a depth

src/flat.mjs
```diff
@@ -1,29 +1,39 @@
...
       }
+    } else {
+      yield iterable
     }
   }
+  return _flat(0, ensureIterable(iterable))
 }
 
-export default function curriedFlat (depth, iterable) {
+export default function curriedFlat () {
```
KSXGitHub: I see you're using `arguments` variable which was never recommended in JavaScript.
A better alternative would be rest parameters.

```javascript
export default function curriedFlat (...args)
```

KSXGitHub:
@sithmel 

> @KSXGitHub can you please have a look at the TS signature? you can now use a function instead of a depth

I had a look at that, and it isn't correct.

BTW, I already created https://github.com/sithmel/iter-tools/pull/81 for `flat(iterable)` and `flat(depth, iterable)`. I will correct your type definition of `flat(fn, iterable)` when this pull is merged (so that I don't have to change my type definition every time you add changes to the code)

KSXGitHub:
@conartist6 I would prefer string not to be flatten

src/flat.mjs
sithmel: Rest args are nicer and more readable. Ok

KSXGitHub:
When will you guys merge this? I have a project that uses `R.flatten()` right now and I would be happy to switch to this.

sithmel:
@KSXGitHub I will merge tomorrow, unless there is some objection

src/flat.mjs
```diff
@@ -0,0 +1,39 @@
...
+  if (typeof depth === 'number') {
+    return (currentDepth, iter) =>
+      currentDepth <= depth && typeof iter[Symbol.iterator] === 'function' && typeof iter !== 'string'
+  }
+  throw new Error('flat: "depth" can be a function or a number')
+}
+
+function flat (shouldIFlat, iterable) {
+  function * _flat (currentDepth, iterable) {
+    if (shouldIFlat(currentDepth, iterable)) {
```
KSXGitHub: I think you should remove iterable check [here](https://github.com/sithmel/iter-tools/blob/0b5d86e12d1950bac9599bd59a9c83992481e834/src/flat.mjs#L9) and move to this line, right before calling `shouldIFlat()` (`typeof iter[Symbol.iterator] === 'function' && shouldIFlat()`).

The current implementation burdens user to check if `iter` parameter is actually iterable (if the user doesn't, `flat()` will attempt to iterate on non-iterables causing errors).

PR #80, opened 2018-12-05
added cursor function

sithmel:


README.md
```diff
@@ -214,6 +215,15 @@ iterable([ 1, 2, 3, 4 ]); // 4, 16
 ```
 This is more memory efficient of using array methods as it doesn't require to build intermediate arrays.
 
+## cursor
+It returns all items with the previous elements returned. Inside an array of a determined size.
```
conartist6: I don't think it should start with undefined. At no point should reasonable code care what the negative second value is('nt).

KSXGitHub:
TypeScript declaration is incorrect BTW.

sithmel:
@KSXGitHub I was going to ask you to have a look, when the api is finalised :smile:

README.md
sithmel: makes sense. FIxed

sithmel:
I think this should be ready to be merged

src/__test__/cursor.test.js
```diff
@@ -0,0 +1,52 @@
...
+  })
+
+  it('frames iterable (cursor equal to the sequence)', async function () {
+    const iter = asyncCursor(5, [1, 2, 3, 4, 5])
+    expect(await asyncToArray(iter)).toEqual([
+      [1, 2, 3, 4, 5]
+    ])
+  })
+
+  it.only('frames iterable (cursor bigger than the sequence)', async function () {
```
conartist6: Oops.

src/__test__/cursor.test.js
```diff
@@ -0,0 +1,52 @@
...
+  })
+
+  it('frames iterable (cursor equal to the sequence)', function () {
+    const iter = cursor(5, [1, 2, 3, 4, 5])
+    expect(Array.from(iter)).toEqual([
+      [1, 2, 3, 4, 5]
+    ])
+  })
+
+  it.only('frames iterable (cursor bigger than the sequence)', function () {
```
conartist6: here too

src/__test__/cursor.test.js
```diff
@@ -0,0 +1,52 @@
...
+    const iter = cursor(6, [1, 2, 3, 4, 5])
+    expect(Array.from(iter)).toEqual([
+      [undefined, 1, 2, 3, 4, 5]
```
conartist6: Again I think it should just be a smaller frame. Or, if you're going to fill values, I think you should allow the user to specify what to fill them with.

sithmel: I don't think that using a smaller frame is an option. I imagine this to be used like this:
```js
for (const [previous, current] of cursor(2, iterable)) {
  ...
}
```
And you can easily manage the undefined value case by case. Just like we do in zip-longest

conartist6: I agree that that is a fairly compelling use case, but if so then you've already hurt it by removing the leading undefineds (at my request). 

I guess I'm starting to think that there are really two methods here: leading cursor and trailing cursor. And I think in both cases current needs to be every item in the list during a cursor run. But that way the pattern would be:

```js
for (const [previous, current] of trailingCursor(2, iterable)) {
  ...
}
```
and 
```js
for (const [current, next] of leadingCursor(2, iterable)) {
  ...
}
```

sithmel: I like the idea. I implemented an option "trailing" to alternate between the 2 variations

sithmel:
@conartist6 are you happy with the current implementation ?

PR #81, opened 2018-12-06
Fix and improve type declaration of flat() and asyncFlat()

KSXGitHub:
## Fix

While `flat()` and `asyncFlat()` might expand any iterable elements found within `iterable` leads to types of individual elements not being preserved, previous type declarations of `flat()` by @sithmel assume the corresponding function preserves type of individual elements; Said declaration is, therefore, incorrect.

The simplest correct declaration should be:

```typescript
declare function flat (depth: number, iter: Iterable<any>): IterableIterator<any>
```

## Improvement

Although `any` is a correct type, it is not really helpful. This pull request also adds a number of overloaded declarations for literal `depth`s within range of 0 to 31 (a.k.a. `ReasonableNumber`).

```typescript
// Type of iter should be IterableIterator<0 | 1 | [2]>
const iter = flat(1, [0, [1], [[2]]] as [0, [1], [[2]])
```

## Note

For ease of diff, this PR targets https://github.com/sithmel/iter-tools/pull/79 instead of master


sithmel:
Thanks @KSXGitHub ! I invited you as a collaborator, happy if you join and if you want to add yourself to the package.json please do

Issue #87, opened 2018-12-16
asyncSlice type signature returns iterable of number


coder-mike:
The type signature for `asyncSlice` returns `AsyncIterableIterator<number>` instead of the expected `AsyncIterableIterator<T>`
https://github.com/sithmel/iter-tools/blob/master/src/index.d.ts#L445

sithmel:
Thanks for spotting it. Merged it.
I am going to release a minor release including this fix and a couple of new features very soon.

PR #88, opened 2018-12-16
Fix return type of slice() and asyncSlice()

KSXGitHub:
Fix https://github.com/sithmel/iter-tools/issues/87


sithmel:
@KSXGitHub thank you so much for the fix!

PR #90, opened 2018-12-16
Add interpose method

conartist6:
Name is taken from same method in Immutable.js.


sithmel:
You can get that easily with:
```
flat(zip(iterable,repeat(null)))
```
Am I right ?

conartist6:
Errr, aside from being three functions instead of one, I think the key difference would be that this only inserts the given item between, not before the first or after the last element.

conartist6:
Yeah. Basically the use case is the same as `Array.prototype.join`, just that you end up with an array instead of a string. This is particularly useful when working with React, where you may have an array of React Elements and want to put some divider (like ", ") between each of them. I'll merge it.

conartist6:
@sithmel could you do another beta release? I added this because I need it right away in sequins.

Issue #95, opened 2018-12-28
async functions and should take an async function as argument


sithmel:
async functions such as **asyncMap**, **asyncFilter** etc. should support using an async function as argument. Example:
```js
asyncFilter(async (id) => await checkdb(id), iterable)
```
The implementation should be straightforward in most cases: just adding an "await" and it should work with both sync and async functions.

sithmel:
really excited for this. I think this will really make this library a feasible replacement for streams.

PR #96, opened 2018-12-29
changed function argument to be optionally asynchronous. amended test…

sithmel:
…s and documentation

https://github.com/sithmel/iter-tools/issues/95


src/async-flat.mjs
```diff
@@ -15,7 +15,7 @@ const defaultShouldIFlat = (depth) => {
 
 function asyncFlat (shouldIFlat, iterable) {
   async function * _asyncFlat (currentDepth, iterable) {
-    if (shouldIFlat(currentDepth, iterable)) {
+    if ((await shouldIFlat(currentDepth, iterable))) {
```
KSXGitHub: Why the extra pair of braces? Why not just `if (await ...)`?

PR #97, opened 2018-12-29
Update TypeScript definition

KSXGitHub:
... to match https://github.com/sithmel/iter-tools/pull/96


sithmel:
thanks @KSXGitHub !

PR #99, opened 2018-12-29
Fix return types of asyncReduce, asyncSome, asyncEvery

KSXGitHub:
**Then:** They return bare values (synchronously)
**Now:** They return promises
**Note:** This pull request includes changes from https://github.com/sithmel/iter-tools/pull/97 and https://github.com/sithmel/iter-tools/pull/96

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,7 +138,9 @@ import { takeWhile } from 'iter-tools/es2018'; // ES2018
 Note: **file names are all lowercase, dash separated. Module names are camelcase.**
 
 ## Async iterators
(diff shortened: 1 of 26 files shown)


sithmel:
I merged #97 instead

KSXGitHub:
@sithmel  Please reconsider, https://github.com/sithmel/iter-tools/pull/97 does not have the fixes. You can tell from the diff and commit history.

KSXGitHub:
@sithmel Specifically, https://github.com/sithmel/iter-tools/pull/99/commits/1cb21d87845d078ec365d6f6a0c3de2074af1979 is the fixes

PR #102, opened 2018-12-30
Enable filter() and find() to narrow return types

KSXGitHub:
Reference: https://github.com/Microsoft/TypeScript/blob/master/lib/lib.es5.d.ts#L1151

### DIFF
--- a/src/__spec__/filter.spec.ts
+++ b/src/__spec__/filter.spec.ts
@@ -1,0 +1,40 @@
+import assert from 'static-type-assert'
+import * as iter from '../index'
+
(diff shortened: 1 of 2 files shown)


sithmel:
thanks @KSXGitHub I'll have a look and merge asap

KSXGitHub:
> @KSXGitHub this seems to be a way for TS to get out the hole it dug itself. But I see the point

Well, TypeScript has many design flaws. They don't intend to make a [sound type system](https://github.com/Microsoft/TypeScript/wiki/TypeScript-Design-Goals) either.

PR #108, opened 2019-01-07
added combination/permutation/etc length

sithmel:
Combinatory generator functions are now iterable objects. With a "length" property method.

I didn't figure out yet how to fix TS declaration, but I am sure it is broken (I tested it)


KSXGitHub:
As long as `get length` is `O(1)`, I'm fine with it.

> I didn't figure out yet how to fix TS declaration, but I am sure it is broken (I tested it)

I may update it soon, but for the time being, use `<ExistingReturnType> & { readonly length: number }`

sithmel:
I added the type declaration as advised

KSXGitHub:
@sithmel Also, since it no longer produce an iterator, but instead, returns iterable instances of some classes, you should also change `IteratorIterable<...> & ...` to `Iterable<...> & ...`

KSXGitHub:
Another thing, is classes necessary? Why don't you just use objects and scoped local functions in place of classes and methods? For example:

```javascript
// instead of
class Instance {
  constructor (value) { this.value = value }
  * [Symbol.iterator] () { yield * this.value }
  get length () { return value.length }
}

// use
const createInstance = value => ({
  * [Symbol.iterator] () { yield * value },
  get length () { return value.length }
})
```

sithmel:
@KSXGitHub why do you think it is better avoiding using classes in this case ?

KSXGitHub:
@sithmel Smaller size, less overhead.

sithmel:
I am now using simple objects

PR #112, opened 2019-01-08
parallel map/filter

sithmel:
This implements an additional (optional) parameter for async-map and async-filter that allows to do operations concurrently.

This is currently missing:
- documentation
- tests
- type declarations

The reason why I opted for this rather then building two new functions are:
* the function returns basically the same values, so it was difficult and counterproductive using a different names
* the implementation for "concurrency=1" is equivalent to the plain asyncMap/asyncFilter

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
  -  async function can take an async callback
  -  find, asyncFind: they return undefined when an item has not been found
 
+### Changed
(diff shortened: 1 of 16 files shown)


sithmel:
I am missing only the type declaration here

sithmel:
@conartist6 in your example you are always waiting for transform to be fulfilled:
```js
    ...
    const ret = await transform(); // you are always waiting here!
    ...
```
The point of this change is the opposite. Mapping by default is sequential, the change enables to run multiple "transform" at the same time.

PR #117, opened 2019-01-15
Fix: iterable() should always return an iterator

KSXGitHub:


sithmel:
I don't understand.
The purpose of the function is to return an iterable.
If an object has a property [Symbol.iterator], it is an iterable.

KSXGitHub:
But the corresponding TypeScript definition return an `IterableIterator`? Hmm, perhaps it is the TypeScript definition that needed to be fixed.

sithmel:
If you need an iterator, just extract it with ```iterable[Symbol.iterator]``` we do that in the codebase every time we want to use ```.next()```

KSXGitHub:
Yep, TS definition is needed to be fixed.

BTW, if you have some kind of handy function that convert both iterable and iterator to an iterator, that would be useful.

```javascript
function * toIterator (iter) {
  yield * iter
}
```

PR #118, opened 2019-01-15
Add partition function

KSXGitHub:
This is a prototype, further improvement might be needed.

**TODO:**
  - [x] Prototype
  - [x] Tests
  - [x] Refactoring
  - [x] TypeScript definition
  - [x] README, CHANGELOG

**DISCUSSION:**
  - Is the tests sufficient?
  - Should we use linked-list instead of array/cursor to store items?
  - Is an async version doable?


sithmel:
If you have a look at the "tee" implementation, I am using a double ended queue, and every time I yield an item, I remove that from the queue.
That way, you can also get rid of the cursors

KSXGitHub:
> If you have a look at the "tee" implementation, I am using a double ended queue, and every time I yield an item, I remove that from the queue.
> That way, you can also get rid of the cursors

So is this supposed to taking advantage of Garbage Collector?

sithmel:
If both iterable are consumed at the same time, they are never filling up the arrays.
For example:
```js
const [evens, odds] = partition(isEven, iterable)
for (const [odd, even] of zip(odds, evens)) {
  ...
}
```

KSXGitHub:
@sithmel I might try to use double ended queue in the future ("Refactoring" section of "TODO" list), but for now, let use settle about tests. If you think some tests are needed, please tell me.

sithmel:
An async version would be great as well. But I suggest to finish off this one first

KSXGitHub:
@sithmel It is done. Please take a look.

src/partition.mjs
```diff
@@ -0,0 +1,42 @@
+import Dequeue from 'dequeue'
+import iterable from './iterable'
+
+const UNDONE = value => ({ done: false, value })
+const DONE = { done: true, value: undefined }
```
sithmel: a matter of style. I found more readable if you just remove these constants and be more explicit

KSXGitHub: But then [these lines](https://github.com/sithmel/iter-tools/blob/ea733dd5cf86d508cdedeb6fd44b9e4047271064/src/partition.mjs#L26-L27) would be ugly.

src/partition.mjs
```diff
@@ -0,0 +1,42 @@
...
+      chosen.push(value)
+    }
+  }
+
+  function part (queue) {
+    function next () {
+      if (queue.length) return UNDONE(queue.shift())
+      if (exhausted) return DONE
+      add()
+      return next()
```
sithmel: can you replace the recursive function with a simpler for loop ?

src/partition.mjs
sithmel: They would be explicit and obvious :smile:

KSXGitHub: I guess it doesn't matter since I'm going to replace the whole `part()` function with a for-loop anyway.

src/partition.mjs
sithmel: Something like:
```js
while (true) {
     if (queue.length) return { done: false, value: queue.shift() }
      if (exhausted) return { done: true, value: undefined }
     add()
}
```

KSXGitHub: I'm curious, what's wrong with recursive function? I find it more readable.

sithmel: First reason is just a matter of style. I found the while true better. But I understand this is subjective.

The second reason is about using the stack. The recursive version insert a lot of functions in the stack and this adds overhead (not big one but still ...). Also the stack have a limited amount of levels http://2ality.com/2014/04/call-stack-size.html

That is solved by tail call optimisation, but it is not widespread yet https://kangax.github.io/compat-table/es6/

So probably not a big deal, but given that the change is minimal ...

sithmel:
Apart  couple of minor changes in the code, it looks very cool!

For me these are missing:
- the async version
- update README
- update CHANGELOG

src/partition.mjs
sithmel: I don't know, but here the change is super simple

KSXGitHub:
> * the async version

I'd like to make a separate pull request after this is settled.



> * update README
> * update CHANGELOG

Added to the TODO list

KSXGitHub:
@sithmel I have updated README and CHANGELOG, the commits is here so even if you choose one of the alternate implementations, you still need to merge this (don't worry about git conflict).

As for the async version, it should be easy to make, however, to be consistent in coding style, I rather wait until you choose one of the implementations.

sithmel:
My favourite is #123. Just to be sure, what should I merge then ?

PR #123, opened 2019-01-17
Alternate implementation of partition() function

KSXGitHub:
Differences from https://github.com/sithmel/iter-tools/pull/118:
  - Uses loop in place for recursion
  - Returns pair of iterators


KSXGitHub:
I tried to change `while (true)` to `while (!exhausted)` to make the code looks nicer but it didn't pass the test. @sithmel, do you have any idea how to improve this?

sithmel:
```while (!exhausted)``` is not correct because you have exhausted the input iterable, but you still need to flush the queue

sithmel:
I like this implementation.

sithmel:
Even better. Shall I close #118 ?

KSXGitHub:
> Even better. Shall I close #118 ?

My TODO list is there!

KSXGitHub:
> Even better. Shall I close #118 ?

Beside, if you merge this, GitHub will automatically close #118

PR #125, opened 2019-01-17
adding support to return method, some minor change for consistency

sithmel:
Apart some tiny change, here are the bigger changes:
1 - changed function names for consistency and to make it easier to debug (*uncurried* is a bit anonymous when you find it in the stack trace)
2 - use *ensureIterable* instead of *iterable*. In an early discussion it was decided that the function takes an iterable as argument. Take into consideration that a generator object is also iterable
3 - whenever we are not using the for..of loop to consume an iterable, we need to deal with the disposal of the original iterator. More info [here](https://github.com/sithmel/iter-tools/wiki/Generator-functions-work-flow)

### DIFF
--- a/src/partition.mjs
+++ b/src/partition.mjs
@@ -1,30 +1,39 @@
 import Dequeue from 'dequeue'
-import iterable from './iterable'
+import ensureIterable from './internal/ensure-iterable'
 
-function uncurried (func, iter) {
+function partition (func, iter) {
   const satisfied = new Dequeue()
   const unsatisfied = new Dequeue()
-  const iterator = iterable(iter)[Symbol.iterator]()
+  const iterator = ensureIterable(iter)[Symbol.iterator]()
+  let exhausted = 0
 
   function * part (queue) {
-    while (true) {
-      while (queue.length) {
-        yield queue.shift()
+    try {
+      while (true) {
+        while (queue.length) {
+          yield queue.shift()
+        }
+
+        const { value, done } = iterator.next()
+        if (done) break
+
+        const chosen = func(value) ? satisfied : unsatisfied
+        chosen.push(value)
       }
-
-      const { value, done } = iterator.next()
-      if (done) break
-
-      const chosen = func(value) ? satisfied : unsatisfied
-      chosen.push(value)
+    } finally {
+      exhausted++
+      if (exhausted === 2) {
+        if (typeof iterator.return === 'function') iterator.return()
+      }
     }
   }
 
   return [part(satisfied), part(unsatisfied)]
 }
 
-const curried = func => iter => uncurried(func, iter)
-
-export default function partition (func, iter) {
-  return iter === undefined ? curried(func) : uncurried(func, iter)
+export default function curriedPartition (func, iter) {
+  if (typeof iter === 'undefined') {
+    return iter => partition(func, iter)
+  }
+  return partition(func, iter)
 }


sithmel:
I have questions too:
why didn't you define a "partition.d.ts" ?
are these files really necessary ?

KSXGitHub:
> why didn't you define a "partition.d.ts" ?

I forgot. 😄

> are these files really necessary ?

It is if user want to import the file directly, i.e.

```typescript
import partition from 'iter-tools/partition'
```

src/partition.mjs
```diff
@@ -1,30 +1,39 @@
...
-      const chosen = func(value) ? satisfied : unsatisfied
-      chosen.push(value)
+        const chosen = func(value) ? satisfied : unsatisfied
+        chosen.push(value)
+      }
+    } finally {
+      exhausted++
+      if (exhausted === 2) {
+        if (typeof iterator.return === 'function') iterator.return()
+      }
```
KSXGitHub: I don't understand, what is this for?

KSXGitHub: Nevermind, I shoud've read [this](https://github.com/sithmel/iter-tools/wiki/Generator-functions-work-flow), my apologies.

src/partition.mjs
```diff
@@ -1,30 +1,39 @@
...
 
   return [part(satisfied), part(unsatisfied)]
 }
 
-const curried = func => iter => uncurried(func, iter)
-
-export default function partition (func, iter) {
-  return iter === undefined ? curried(func) : uncurried(func, iter)
+export default function curriedPartition (func, iter) {
+  if (typeof iter === 'undefined') {
```
KSXGitHub: What's the difference between `x === undefined` and `typeof x === 'undefined'`? I think `x === undefined` should be faster since it does not deal with string comparison 🤔.

sithmel: It is an old js baggage. Due to the fact that you can sadly redefine undefined.
But I can see now, that is not recommended anymore http://2ality.com/2013/04/check-undefined.html
I will reverse that

src/partition.mjs
sithmel: iterator.return is only useful to close an iterator **if is not exhausted**

sithmel: More info here: http://exploringjs.com/es6/ch_iteration.html#_closing-iterators-via-return

src/partition.mjs
KSXGitHub: > I will reverse that

No needs if you want to stay consistent.

But if you really want to use `undefined`, I think you should also make similar changes in other places.

src/partition.mjs
KSXGitHub: Man! This iterator thing is more complicated than meets the eyes. Anyway, is there a way to make a wrapper that calls `iter.return()`? I prefer a clean and uncluttered code.

src/partition.mjs
sithmel: I'll review the code base afterwards

src/partition.mjs
sithmel: I don't think there is a neater way. The good news is that is only necessary when you are consuming an iterator manually using "next". In any other case this is automatic.

KSXGitHub: > I don't think there is a neater way. 

What about https://github.com/sithmel/iter-tools/pull/129?

I can go even further, make a wrapper that also support `tee()`.

KSXGitHub:
@sithmel It's been a few days, will you merge this? Also consider https://github.com/sithmel/iter-tools/pull/129.

sithmel:
@KSXGitHub I propose I start merging this, then we add asyncPartition.
Then we merge #129 depending on our discussion about multiPartition.

Is that ok ?

KSXGitHub:
@sithmel That's exactly what I wanted.

Issue #134, opened 2019-01-18
Suggestion: function that split iterable into two at a given position


KSXGitHub:
## Split by Index

```javascript
function splitAt (position, iter) {
  const [first, second] = tee(iter)
  return [
    slice({ end: position }, first),
    slice({ start: position }, second)
  ]
}
```

## Split by Condition

...

conartist6:
See my comments on https://github.com/sithmel/iter-tools/pull/135, I think they would also apply to this function.

conartist6:
We shouldn't be helping people do things with iterables that they shouldn't be using iterables to do.

KSXGitHub:
@conartist6 But won't be as performant

sithmel:
I would move the discussion to #136 that seems to be fulfilling the same goal

KSXGitHub:
> I would move the discussion to #136 that seems to be fulfilling the same goal

You're definitely right.

KSXGitHub:
Reopen as https://github.com/sithmel/iter-tools/issues/136#issuecomment-455817799 arises.

conartist6:
Yeah having had these discussions I like your `splitAt` proposal better now.

sithmel:
This is blocked until *fork* is ready #151

sithmel:
fork is now on master @KSXGitHub 

I like both splitAt and splitBy ideas ;-)

PR #138, opened 2019-01-20
Create splitAt() and asyncSplitAt()

KSXGitHub:
Fixes https://github.com/sithmel/iter-tools/issues/134

**Discuss:**

 * [x] What should `asyncSplitAt()` return?
  - a tuple of 2 async iterators
  - ~~a tuple of 2 promises of arrays~~
  - ~~a promise of a tuple of 2 arrays~~


sithmel:
>What should asyncSplitAt() return?
an array with 2 async iterators like you did, is ok to me

src/split-at.mjs
```diff
@@ -0,0 +1,18 @@
+import slice from './slice'
+import tee from './tee'
+import ensureIterable from './internal/ensure-iterable'
+
+function splitAt (position, iter) {
+  const [a0, b0] = tee(ensureIterable(iter))
+  const a1 = slice({ end: position }, a0)
```
conartist6: I would like to see a better implementation. In particular I want to see that if the user consumes the entire first half of the split prior to consuming any of the second, that no cache values are created. If the user begins consuming the second half of the split prior to the first, I want to see and array cache of the correct size created (`new Array(position)`) and filled by calling `next()` `position` times, which must be done by hand in order to avoid closing the source iterable accidentally. This means that you will also need a `try..finally` block to ensure that `iter.return()` is called if `splitAt` returns prior to exhausting `iter`.

KSXGitHub:
Close in favor of https://github.com/sithmel/iter-tools/pull/181

Issue #137, opened 2019-01-20
Suggestion: pipe() — like pipeline(), but does not return a function


KSXGitHub:
## Why?

Since `pipeline` is just a reverse `compose`, I must use 2 pairs of parentheses (2 function calls) every time I use `pipeline`. That's not only makes the code looks ugly, it's also annoying.

## Suggestion

Create a `pipe` function that unlike `pipeline`, it takes 1 value and many functions and returns a value.

```javascript
const result = pipe(
  range(),
  filter(x => x % 2 === 0),
  map(x => x % 3)
)
```

sithmel:
You can do that:
```js
const result = pipeline(
  () => range(),
  filter(x => x % 2 === 0),
  map(x => x % 3)
)
```

KSXGitHub:
@sithmel You missed the point: `result` in your example is a function whilst in mine, it is an iterable.

So your example should rather be:

```javascript
const result = pipeline(
  () => range(),
  filter(x => x % 2 === 0),
  map(x => x % 3)
)() // ← extra pair of parentheses
```

sithmel:
my bad. let's fix pipeline instead!

KSXGitHub:
BTW, The implementation of `pipe` is rather simple:

```javascript
function pipe (x, ...fns) {
  return fns.reduce((x, fn) => fn(x), x)
}
```

as opposed to `pipeline`:

```javascript
function pipeline (...fns) {
  return fns.reverse().reduce((f, g) => (...args) => f(g(...args)))
}
```

KSXGitHub:
> my bad. let's fix pipeline instead!

Would that be a breaking change?

sithmel:
nah, I never properly released that.

As a matter of fact it is about time to release a new version

KSXGitHub:
You can use my `pipe` implementation above.

And please don't release before I update its type definition.

PR #140, opened 2019-01-20
fix to pipeline

sithmel:
see #137 
Type declaration is missing


src/pipeline.mjs
```diff
@@ -1,3 +1,3 @@
-export default function pipeline (...fns) {
-  return fns.reverse().reduce((f, g) => (...args) => f(g(...args)))
+export default function pipeline (iterable, ...fns) {
```
KSXGitHub: About the argument name "iterable": While users of `iter-tools` mostly focus on iterables, this function is applicable to other types as well, the following code is valid for example:

```javascript
const result = pipeline(
  3,
  x => x * 2,
  x => x + 1,
  x => x * x
)
```

KSXGitHub:
Should we rename the function to `pipe`? It is shorter, it is also more familiar as other FP libraries provide `pipe`.

src/__test__/pipeline.test.js
```diff
@@ -11,12 +11,7 @@ function isEven (x) {
 
 describe('pipeline', function () {
   it('using function', function () {
```
KSXGitHub: This line is not only no longer necessary, it is somewhat misleading

sithmel:
@KSXGitHub please fix the type declaration. I'll take care of the rest and to merge

KSXGitHub:
> @KSXGitHub please fix the type declaration. I'll take care of the rest and to merge

https://github.com/sithmel/iter-tools/pull/142

(How did you miss that?)

KSXGitHub:
BTW, https://github.com/sithmel/iter-tools/pull/142 assumes that you use the name `pipeline` just in case you want to rename the function to `pipe`.

Issue #151, opened 2019-01-22
functions to use in a pipeline


sithmel:
In this ticket, I'd like to discuss 2 new functions and a couple of changes. But before, let me explain the context.
Last summer I've experimented with data pipelines, using async-iterables instead of streams. Using data pipelines you have often the necessity to fork the iterable and to apply different sub-pipelines to the forked ones. Here's an example:
```js
...
const [copy1, copy2] = asyncTee(iterable, 2)
const transformed1 = aPipeline(copy1)
const transformed2 = anotherPipeline(copy2)
const merged = asyncMerge(asyncMergeByReadiness, [transformed1, transformed2])
for await (const item of merged) {
  // ...
}
```
I'd like the previous code to be simpler. So here's a list of proposal:

## 1 - merge - asyncMerge default function
merge - asyncMerge currently requires a function to decide what item consume first. Adding a sensible default can simplify most common cases. I propose to use a round robin algorithm on merge and asyncMergeByReadiness on asyncMerge.

## 2 - fork: a better tee
tee api come directly from the [Python itertools](https://docs.python.org/2/library/itertools.html#itertools.tee). It would be very convenient having a curriable version. Even better if it returns a iterable like @KSXGitHub did on the [multiPartition](https://github.com/sithmel/iter-tools/issues/116#issuecomment-455700232), so we don't need to specify the number of iterables returned. Something like:
```js
const [copy1, copy2] = asyncFork(iterable)
// equivalent to
const [copy1, copy2] = asyncTee(iterable, 2)
```

## 3 - apply multiple pipelines to an array of iterables
I could not come up with a good name yet. Let's call this function pipeMultiple for now.
```js
const [transformed1, transformed2] = pipeMultiple([aPipeline, anotherPipeline], [copy1, copy2])
// equivalent to
const transformed1 = aPipeline(copy1)
const transformed2 = anotherPipeline(copy2)
```
It should also be curriable.

## the result
```js
...
const merged = pipe(
  asyncFork(),
  pipeMultiple([aPipeline, anotherPipeline]),
  asyncMerge()
)

for await (const item of merged) {
  // ...
}
```

KSXGitHub:
> ## 2 - fork: a better tee
[...]

PR #153, opened 2019-01-23
Create fork()

KSXGitHub:
**Other changes:**
  - Installed advanced-map-initialized

**TODOs:**
  - [x] Working prototype
  - [x] TypeScript definition
  - [ ] **Discuss:** How to handle iterator.return(), if at all?
  - [ ] Add more tests
  - [ ] Async versions
  - [ ] README and CHANGELOG

**Related issue:**
  - https://github.com/sithmel/iter-tools/issues/151

### DIFF
--- a/src/fork.mjs
+++ b/src/fork.mjs
@@ -1,9 +1,6 @@
-import AdvancedMapInitialized from 'advanced-map-initialized'
-import map from './map'
-import range from './range'
 import ensureIterable from './internal/ensure-iterable'
 
-export default function fork (iterable) {
+export default function * fork (iterable) {
   const queue = []
   const iterator = ensureIterable(iterable)[Symbol.iterator]()
 
@@ -23,10 +20,7 @@ export default function fork (iterable) {
     }
   }
 
-  const resultMap = new AdvancedMapInitialized(Map, part)
-
-  return map(
-    index => resultMap.get(index),
-    range()
-  )
+  while (true) {
+    yield part()
+  }
 }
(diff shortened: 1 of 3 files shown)


sithmel:
> How to handle iterator.return(), if at all?

this needs to be handled.

I thought we could take note of all generator instances we start ... and then call ```.return``` on the input iterable when all of them are exhausted ...  but I realised this is not correct.

When I exhaust all opened generators, then I am unable to start using a new one.

src/fork.mjs
```diff
@@ -0,0 +1,32 @@
...
+    while (true) {
+      while (cursor < queue.length) {
+        yield queue[cursor]
+        cursor += 1
+      }
+
+      const { value, done } = iterator.next()
+      if (done) break
+
+      queue.push(value)
```
conartist6: Iterables *should* return the same value every time `next()` is called after they are exhausted, but some may not. It is a good idea to capture the first done value explicitly to avoid bugs when dealing with user defined iterators/iterables.

src/fork.mjs
```diff
@@ -0,0 +1,32 @@
...
+    }
+  }
+
+  const resultMap = new AdvancedMapInitialized(Map, part)
```
conartist6: I don't see a reason that a map of any kind is needed. Seems like a solution searching for a problem. Instead I think you simply mean to write:
```js
export default function * fork() {
  function * part() {
  }
  while(true) {
    yield part();
  }
}
```

sithmel:
@KSXGitHub we are converging on the implementation on #161 .
can I close this one ?

KSXGitHub:
Sure

PR #155, opened 2019-01-23
fork() that handles iterator.return() 

KSXGitHub:
**Limitation:** iterator.return() is supposed to called when all sub-iterators are exhausted,
but since there's no limit to number of sub-iterators, iterator.return() will be called
when all accessed sub-iterators are exhausted.


conartist6:
Actually I think your example and code are not quite correct. The first example should be:
```js
const [a, b, c] = fork(iterator)
Array.from(a) // iterator.return should be called right after this line
Array.from(b)
Array.from(c) // instead of this one
```

sithmel:
Interesting experiment! but the limitation is not worth it IMO

sithmel:
I would accept that we have to set the number of iterables returned

conartist6:
@sithmel But my point is that here and in tee, you can safely close the source iterable as soon as *any* resulting iterable is exhausted, because at that point you know that you have all the values in your cache(s). Then there is no limitation.

sithmel:
@conartist6 
> you can safely close the source iterable as soon as any resulting iterable is exhausted

I don't think you can.

If you are slicing and consuming the first iterable, you might still want to consume the whole second iterable.

conartist6:
But you're forgetting that you still can consume the whole second iterable. You were only going to read its values from the cache anyway since you already exhausted the source.

conartist6:
So if you use slice, that touches on what I brought up in https://github.com/sithmel/iter-tools/issues/156. But really the implementation shouldn't have any trouble distinguishing an early return from the iterable being exhausted.

The problem becauses that if all the resulting iterables are run through slice, then you'd never close the source.

I think that's fine though. All that needs to happen is that the `fork` iterable (the one which yields the cached copies of the source iterable) needs to have a return method. When you call return on that iterable you are indicating that you no longer need to create any more forks, and that when all the existing forks have returned the source iterable return method should be called, even if the source iterable is not exhausted.

sithmel:
@conartist6 
Let's say I call:
```js
const [copy1, copy2] = fork(iterable)
```
and then you consume the first copy:
```js
array.from(slice(4, copy1)) 
```
Using array.from will consume the iterable returned by slice. Slice will automatically call the "return" method. This is because *slice* uses "for..of" and that will call *return* as soon as it finishes the loop (or call break/return).

Given that we didn't even start consuming "copy2", we can't be aware that we want to consume another instance.

conartist6:
Yeah but that's not a problem. Slice calls return, but the fork impl knows the difference between a consumer calling return and a consumer the source iterable being exhausted. It does not call return on the source iterable unless a consumer exhausts the source.

conartist6:
Take this example:
```js
const [a,b,c] = fork(infiniteIterable); // destructuring calls return on fork. Fork now knows it will only ever create three iterables!
Array.from(slice(0, 3, a)); // slice calls return here, but source not exahausted; not closed
Array.from(slice(0, 3, b)); // same as above
Array.from(slice(0, 3, c)); // iterable is closed here, because all forked iterables that have or will be generated have called return.
```

conartist6:
Even if you're creating an arbitrary number of iterables using `slice(n, fork())`, slice calls return on fork allowing it to know when there's no more work to do. Only in the case that somebody does `fork()[Symbol.iterator].next()` need we worry that they shall cause a problem by forgetting to call `return()`, and I can't imagine any situation with a legitimate need to do that. If we've learned anything writing this library it's that handling raw iterators is a complicated enough task that you should pretty much always delegate it to a library.

conartist6:
No, it will work in that case. How about this: I'll just write it and my tests will prove that all these cases work.

conartist6:
Actually the example at the top of this issue is illegal:
```js
const container = fork(iterator)
const abc = slice({ start: 0, end: 3 }, container)
const def = slice({ start: 3, end: 6 }, container) // can't slice an iterable twice without forking it

// consume all of abc
Array.from(abc).map(iter => Array.from(iter)) // iterator.return is called right after this line despite def is still not exhausted

// consume all of def
Array.from(def).map(iter => Array.from(iter)) // undefined behavior
```

conartist6:
[Here](https://github.com/sithmel/iter-tools/blob/fork/src/fork.mjs) is the implementation I discussed. Just wanted to share it before I head to work. I'll write the tests for it next.

conartist6:
Uh, nope, not correct? Firstly the for loop you've written will run forever. You don't actually want to create infinitely many copies of the iterable.

And secondly, the second time through your loop the iterable *is* closed already. You've said you consume it, presumably fully with `Array.from` or similar. If it is fully consumed it there's no other logic necessary. You don't call `return()` when you exhaust an iterable, it simply tells you what its return value is. My code handles that.

sithmel:
```js
const iterables = fork(iterable)
for (const [n, iter] of enumerate(iterables)) {
  console.log(Array.from(slice(3, iter)))
  if (n===3) break
}
```
this ?

conartist6:
Nice try! That does actually break my code, though not the concept; the fix is trivial. Our `zip` implementation correctly calls return on its sources. Enumerate doesn't create any iterator of its own and so is a non-factor. `break` (or `return`, or an exception, before you get clever) inside `for..of` over an iterable is defined in the language spec to close the iterable. Slice iterables close their source iterables when the sliced portion is consumed.

So the parts go together: the `queue` array holds three items with the source iterable open, and this is the state that the code is in up until the `break`. When `break` happens, the fork generator becomes closed, looks and notices that all fork iterables up to this point have already returned (I forgot to write this), and calls `return()` on the source.

sithmel:
@conartist6
Let me rephrase to be sure I understood:
* let's define **sourceIterable** the input of *fork*
* *fork* returns an iterable of iterables. Let's call it **iterableOfIterables**
* let's call **forkedIterable** every iterable yielded by **iterableOfIterables**

So:
I call *return* on the **sourceIterable** IF:
* *return* has been called on **iterableOfIterables**
AND
* *return* has been called on every **forkedIterable** already returned by **iterableOfIterables**

If this is correct, I finally follow :+1:

conartist6:
This is correct.

sithmel:
@KSXGitHub we are converging on the implementation on #161 .
can I close this one ?

Issue #156, opened 2019-01-23
tee should allow its iterators to return early


conartist6:
A selling point of tee the way we wrote it is that it will work on infinite iterables -- as long as the consumers keep consuming each iterable resuling from the tee all the old entries are removed from the dequeues and we don't keep much of anything in cache.

But what if one of those consumers stops consuming? Now its cache will begin to fill up -- essentially creating a memory leak.

We should return an iterable with a `return` method, which method cleans up by destroying that iterable's tee cache.

conartist6:
I also think tee should only keep a single dequeue internally, whose size is determined by the consumer lagging farthest behind the current item in the source iterable.

sithmel:
True. this needs to be addressed

sithmel:
partition has the same problem. If a generator function is exhausted, it will still queue items
```js
function partition (func, iter) {
  const satisfied = new Dequeue()
  const unsatisfied = new Dequeue()
  const iterator = ensureIterable(iter)[Symbol.iterator]()
  let exhausted = 0

  function * part (queue) {
    try {
      while (true) {
        while (queue.length) {
          yield queue.shift()
        }

        const { value, done } = iterator.next()
        if (done) break

        const chosen = func(value) ? satisfied : unsatisfied
        chosen.push(value)
      }
    } finally {
      exhausted++
      if (exhausted === 2) {
        if (typeof iterator.return === 'function') iterator.return()
        // replace satisfied or unsatisfied with a fake object that puts items in /dev/null
      }
    }
  }

  return [part(satisfied), part(unsatisfied)]
}
```

sithmel:
this has been fixed with the new fork

PR #160, opened 2019-01-26
buffer data structure for tee/fork

sithmel:
This should solve the issue https://github.com/sithmel/iter-tools/issues/156 and provides a base to develop *fork* easily.

I have created an intermediate abstraction called MessageQueue. I hate the name and I couldn't come up with a better metaphor. Apart that, I am pretty happy of how it works.

Comments welcome!


sithmel:
When used as a queue, messageQueue performance are equivalent to dequeue (slightly better actually). So it may be worth dropping dequeue as a dependency.

```js
module.exports['dequeue 100000 items'] = function () {
  const q = new Dequeue()
  for (const n of range(10000)) {
    q.push(n)
  }
  while (q.length) {
    q.shift()
  }
}

module.exports['messageQueue 100000 items'] = function () {
  const q = new MessageQueue()
  for (const n of range(10000)) {
    q.add(n)
  }
  while (!q.isExhausted()) {
    q.get()
  }
}
```
```
  Dequeue against MessageQueue
    ✓  dequeue 10000 items x 2,558 ops/sec ±0.45% (92 runs sampled)
    ✓  messageQueue 10000 items x 2,708 ops/sec ±0.27% (94 runs sampled)

    Result: messageQueue 10000 items is 5.87% faster than dequeue 10000 items.
```

conartist6:
Actually I needed to write my own queue for my optimal fork implementation because it needs to be able to swap implementations. I'll post the code when I'm done, probably tonight (California time)

conartist6:
Mine should be more efficient than dequeue as well since it is singly linked.

sithmel:
I have applied the new data structure everywhere and I got rid of dequeue.

I have now implemented 2 different abstractions, based on the same code:

### queue
Good old queue (single linked list) implementing:
* shift
* push
* isEmpty

All in O(1)

## exchange
I have borrowed the name from messaging protocol libraries (https://www.rabbitmq.com/tutorials/amqp-concepts.html#exchange-fanout).
The *exchange* can:
* create a consumer with **spawnConsumer**
* **push** an item in the queue
* use **noMoreConsumers** to declare that no more consumers will be created and therefore consumed items can be garbage collected

A *consumer* implements:
* shift
* isEmpty

So basically, you push items in the exchange and read them with the consumer.
Everything is O(1) here, too.

PR #162, opened 2019-01-27
Fork

conartist6:
This is still a work in progress. I'm having myself a bit of a skiing weekend at the moment so I likely won't be fine tuning it immediately, but I think the optimizations I implemented are worth consideration.

Basically the advantages of this implementation is that it can work for all our use cases. For non-infinite iterables the likelihood is that the user fill exhaust one fork before they touch the other. In this case the implementation is fast, because an Array cache is optimal in that situation.

However if the user uses destructuring syntax for example, which results in closing the iterable of iterables before consumption of any of the forked iterables begins, then this implementation is just as a efficient as tee and can be used for infinite iterables.

It is also more memory efficient when there are a significant number of iterables, because the singly linked list cache has immutable nodes which are shared between all the forks.


conartist6:
The weirdest code in here is for the weirdest case -- where the user creates many forked iterables before the fork generator is closed, then starts consuming the forked iterables, then closes the fork generator. In this case the code does something rather clever to convert the array caches into linked list caches efficiently, but also I'm not sure it wouldn't also be fine to just demand that users choose up front: close the iterable of iterables before you start consuming the forked iterables and you will get linked caches and support for infinite iterables, else you will get array caches.

conartist6:
I am closing this request in favor of https://github.com/sithmel/iter-tools/pull/161.

PR #164, opened 2019-01-27
removing iter and async-iter

sithmel:
In preparation for the new major release

This was supposed to be a trivial task, but one test is now failing (only using ES5).
I have tried to revert and everything works.

If anybody can have a look it would be amazing, or I can have a look as soon as I can.


conartist6:
For one thing, could we eliminate the `forceWrap` parameter to this method and simply make a separate file for `asyncify`, so that you could just import and use that function directly where currently we have `ensureIterable(..., true)`.

conartist6:
I can't imagine quite why removing `iter` had this effect, but the error, once decoded, seems to be that in an Async Generator `yield*` is expecting to be invoked on an async iterable. An `asyncify` implementation that works everywhere would be:
```js
export default async function * asyncify (iterable) {
  if (iterable[Symbol.asyncIterator]) {
    yield * iterable 
  } else {
    for (const item of iterable) {
      yield Promise.resolve(item)
    }
  }
}
```

We should probably attache a comment/TODO to that to figure out if there will be a point in the future when we can safely return it to its original implementation, which I'd imagine is correct since node handles it fine natively.

conartist6:
I have added my suggested changes to the PR.

sithmel:
thanks @conartist6 
I would merge

PR #167, opened 2019-01-30
map func

### LINKED ISSUE #158: mapFunc (temporary name)
pipeMulti carries on multiple iterable transformations in parallel:
```js
const iter1 = range(10)
const iter2 = range(10)
const even = filter((n) => n % 2 === 0)
const square = map((n) => n * n)
const [evenIter1, squareIter2] = pipeMulti([even, square], [iter1, iter2])
```
It can be used in pipelines, leveraging fork (or tee)
```js
const [evenIter1, squareIter2] = pipe(
  range(10),
  fork(2), // this fork return 2 items
  pipeMulti([even, square])
)
```
Given that often times you precede pipeMulti with "fork", it would be nice if pipeMulti can fork automatically whenever the second argument is an iterable instead of an array.
```js
const iter = range(10)
const even = filter((n) => n % 2 === 0)
const square = map((n) => n * n)
const [evenIter1, squareIter2] = pipeMulti([even, square], iter)
```
That become:
```js
const [evenIter1, squareIter2] = pipe(
  range(10),
  pipeMulti([even, square]) // it implicitly fork(2)
)
```

sithmel:
Fixes #158


conartist6:
This is one conversation that I haven't been following up until now, but I admit that this function is a real head scratcher to me. What is it for?

src/map-func.mjs
```diff
@@ -0,0 +1,9 @@
+import curry from './internal/curry'
+import map from './map'
+import zip from './zip'
+
+function mapFunc (functions, inputs) {
+  return map(([func, input]) => func(input), zip(functions, inputs))
```
conartist6: Wouldn't a better implementation be:
```js
merge((item, func) => func(item), [[...items], [...funcs]])
```
And being rather terse (and also fairly descriptive) wouldn't it be better to use this implementation in code should this presumably rather rare need arise?

conartist6: OK, I've just taken a careful look at merge, and it isn't at all what I thought it was.

sithmel:
@conartist6 
When building a data pipeline, you often need to fork an iterable, to transform the forked version in parallel and to merge them together at the end.
```js
...
const dataPipeline = pipe(
  asyncFork,
  mapFunc([aPipeline, anotherPipeline]),
  asyncMerge()
)

for await (const item of dataPipeline(iterable)) {
  // ...
}
```

sithmel:
The current work stream (fork, mapMulti, consumeMany, partition, splitBy etc.) is focused in improving how the library works with iterables of iterables.

conartist6:
I'm still stuck on how we're doing this at all. I'd argue that the example you've given is not good code. You've created two separate paths and managed to avoid naming them. Instead presumably in this future there would be a list of transforms like
```js
const [left, right] = asyncFork(iter);
pipe(
  [left, right],
  mapFunc([leftMapper, rightMapper]),
  filterFunc([leftFilterer, rightFilterer]),
  whateverFunc([leftWhatever, rightWhatever]),
  asyncMerge(),
)
```
But this is crazy! Code should not be written with structure described as things which sit side by side with one another.

It reads like that almost-oo pattern where synchronized array
```js
const firstNames = ["Johnathan", "Holt",      "Anna"]
const lastNames =  ["Groff"    , "McCallany", "Torv"]
// When this becomes long enough that each line wraps the structure of it becomes obscured
```
It looks cute at first, but as soon as you try to use the pattern for anything serious it's going to be really unpleasant trying to pack the code you write into a two dimensional space. 

The example above, then I'd argue, should be written like this:
```js
const [left, right] = asyncFork(iter);
asyncMerge(
  pipe(
    left,
    asyncMap(leftMapper),
    asyncFilter(leftFilterer),
    asyncWhatever(leftWhatever),
  ), pipe(
    right,
    asyncMap(rightMapper),
    asyncFilter(rightFilterer),
    asyncWhatever(rightWhatever),
  )
)
```

Not all code duplication is evil, especially where it is easy to recognize a repeated pattern.

KSXGitHub:
I think naming the function `mapFunc` confuses @conartist6. The reason @sithmel used this name is because "it looks like a reverse version of `map`". But upon further examination, it looks more of a `zip` than a `map`, so I suggest to change the name to something else, such as `matchFunc`.

conartist6:
You'll excuse me if I don't feel that my problem is being confused by the name, though the name could certainly be more descriptive. I think I was clear about what my problem with it is, and it is quite a lot deeper than the name.

sithmel:
The fact that you can use mapFunc to shoot yourself in the foot, doesn't necessarily make it bad.
For example I can rewrite your pipeline like this:
```js
const [left, right] = asyncFork(iter);
const leftMap = pipe(leftMapper, leftFilterer, leftWhatever)
const rightMap = pipe(rightMapper, rightFilterer, rightWhatever)

pipe(
  asyncFork(iter),
  mapFunc([leftMap, rightMap]),
  asyncMerge(),
)
```
And it is perfectly readable. Actually it makes me visualise the data flow better.
I find the alternative a bit more difficult to read:
```js
const [left, right] = asyncFork(iter);
const leftMap = pipe(leftMapper, leftFilterer, leftWhatever)
const rightMap = pipe(rightMapper, rightFilterer, rightWhatever)
asyncMerge([leftMap(left), rightMap(right)]),
```

The only thing that I agree is not great is the name

conartist6:
I don't think the example your using shows the real strengths or weaknesses of either syntax. In order to see that, I think you'd need to replace the names with actual calls to iter-tools methods including callback functions and parameters. It's almost a certainty that any non-toy call will be long enough that two of them will not fit on the same line together inline, and some of them will need to break into multiple lines. This will necessitate naming particular transfom functions, which I think will usually prove difficult. It will also necessitate a non-linear reading, as the coder will keep having to flip back and forth between the names of transforms and their definitions, which are likely to become out of order or interleaved with other helpers or pushed off the top of the page, all things which make reading the code more difficult.

Writing the code becomes more difficult too, as while it is not too hard to pick names for concrete things, it is much harder to name abstract actions that only make sense in some other specific context, such as when interpreted as a callback to some specific function.

Let me repost my example:
```js
const [left, right] = asyncFork(iter);
asyncMerge(
  pipe(
    left,
    asyncMap((item, index) => ({item, index, isEven: index % 2})),
    asyncFilter(item => typeof item !== 'undefined'),
    asyncWhatever(...),
  ), pipe(
    right,
    asyncMap(...),
    asyncFilter(...),
    asyncWhatever(...),
  )
)
```

Notice how all the things which are most difficult to name are in spots where they don't need to be named, and they all have space to take up a whole line or several if they need to. As such a block grows larger and more complicated, most editors will support simply folding up one of the pipe calls if it is not programmer's focus.

sithmel:
Last summer I worked on a data migration. I repeatedly used this pattern:
* fork
* run some subpipeline in parallel
* zip all together
* fork
* and so on...

Like this:
```js
pipe(
  csvReader(...),
  asyncFork,
  mapFunc([readDataFromMongo(...), readDataFromApi(...), generateNewItem(...)])
  asyncZip,
  asyncFork,
  mapFunc([pipe(asyncMap(...), writeCSV(...)), getAttachment(...)]),
   ...
)
```
This helped me to visualise the sections of the pipeline. What was running concurrently and so on.

sithmel:
What about if I change the function name to 
**multiMap**
or
**mapMulti**

as suggested here? https://github.com/sithmel/iter-tools/issues/158#issuecomment-457219932

I would really like to merge this

KSXGitHub:
> What about if I change the function name to
> **multiMap**
> or
> **mapMulti**

If you plan to make other equivalents (such as `filterFunc`), I suggest to not merge this at all, instead, create a wrapper function (`wrapper(map)` instead of `mapFunc`).

Otherwise, I'm fine either way.

sithmel:
I don't plan to follow with anything else (no async, no filter etc.).
My only use case is to map multiple iterable in multiple functions (see examples above).

sithmel:
I think I have finally found a name for this:
**applyMulti**

From wikipedia:

> In mathematics and computer science, apply is a function that applies functions to arguments.  ...

This applies multiple functions to multiple arguments

conartist6:
I mean javascript has apply, and it works very differently. What about something like `applyThunked` or simply `thunk`? I like that because thunk usually indicates an inversion of control pattern, which this is. Instead of calling a function on an iterable (as with apply) an iterable is calling a function with itself / its items.

sithmel:
From wikipedia:

> In computer programming, a thunk is a subroutine used to inject an additional calculation into another subroutine. Thunks are primarily used to delay a calculation until its result is needed

In lisp apply is defined like this: http://jtra.cz/stuff/lisp/sclr/apply.html

That is not conceptually different from js, apart for the fact that that is a method instead of one of the arguments: func.apply(...args...).

In here I am applying an iterable of arguments to an iterable of functions, hence applyMulti.

conartist6:
Yes, and here we are injecting additional calculation into the next method of the outer iterable.

sithmel:
Formally this is an utility function that has no strict relation with iterables.
It just happened to be very convenient to be used with them. I show this in the README:

## apply-multi
This function takes as argument an iterable of functions and an iterable of arguments. The result will be an iterable with the results of all the functions applied to all the arguments. The iterable will yield a number of items equal to the shorter iterable.
```js
const square = x => x * x
const double = x => x * 2
applyMulti([square, double], [3, 10]) // 9, 20
```
It is particularly useful to map different iterables or async iterables to different transformations:
```js
const [squareEven, squareOdds] = applyMulti([map(square), double(double)], partition(even, iterable))
```

conartist6:
My proposal:
Name: `mergeApply`
Implementation:
`(fns, ...iterables) => merge((f, ...args) => f(i, ...args), fns, ...iterables)`

sithmel:
We can consider that implementation when "merge" is on master.

But the name mergeApply doesn't represent what the function does.

What's wrong with applyMulti ? Ticks all the boxes for me

conartist6:
I'd also recommend that if the first argument is a function not an iterable of functions, we treat it as `execute(() => fn)`.

This gets rid of the red herring of `multi`. This function is perfectly well defined in non-higher-order cases, which is the simpler way to name and understand it anyway.

sithmel:
The function is designed to work in data pipelines as I explain https://github.com/sithmel/iter-tools/pull/167#issuecomment-459056249

conartist6:
Does `mergeApply` not describe exactly what the function does? It does `merge` (as defined in the pending branch) with the merge function being, in essence, `apply`.

conartist6:
I may have confused myself looking at your example. I do find it a little telling that you got the example wrong, since we're talking about something so abstract.
```js
const square = x => x * x
const double = x => x * 2
applyMulti([square, double], [3, 10]) // 9, 20
```
That should be: 
`applyMulti([square, double], [[3], [10]) // [[9], [20]]`

sithmel:
mergeApply:
It **merges** 2 iterables into 1, by **applying** the function in the first iterable to the argument in the second.

applyMulti
it is like **apply** but for **multi**ple pairs function/argument

I prefer applyMulti

This is the up to date documentation:

## apply-multi
This function takes as argument an iterable of functions and an iterable of arguments. The result will be an iterable with the results of all the functions applied to all arguments. The iterable will yield a number of items equal to the shorter iterable.
```js
const square = x => x * x
const double = x => x * 2
applyMulti([square, double], [3, 10]) // 9, 20
```
It is particularly useful to map different iterables or async iterables to different transformations:
```js
const [squareEven, doubleOdds] = applyMulti([map(square), double(double)], partition(isEven, iterable))
```

sithmel:
bear with me there is another problem in the example :man_facepalming:

sithmel:
fixed:

## apply-multi
This function takes as argument an iterable of functions and an iterable of arguments. The result will be an iterable with the results of all the functions applied to all arguments. The iterable will yield a number of items equal to the shorter iterable.
```js
const square = x => x * x
const double = x => x * 2
applyMulti([square, double], [3, 10]) // 9, 20
```
It is particularly useful to map different iterables or async iterables to different transformations:
```js
const [squareEven, doubleOdds] = applyMulti([map(square), map(double)], partition(isEven, iterable))
```

conartist6:
Agh I keep tripping up. Square and double were not methods over iterables.

Let me refine the idea even further.
Let's say we have my `merge` function.
We could write another iter-tools method: `apply` whose implementation was:
```js
apply(fn, iterableOfArguments) // fn(...iterableOfArguments)
```

A very simple function to be sure, but also one that is very simple to understand. In that case we would not have this method at all. The pattern would simply be:

```
import {merge, apply} from 'iter-tools';

merge(apply, functions, arguments)
```

sithmel:
It is a brilliant idea :+1: 

But I would still want a single function because it is something I'll finish using pretty often.

And I would still call it applyMulti :-)

sithmel:
also the way I need this, I won't spread the arguments in the functions. Just passing them as they are. 1 argument for every function.

conartist6:
In your pipeline example, the code now looks like:

```js
pipeExec(
  csvReader(...),
  asyncFork(3),
  merge(apply).bind([readDataFromMongo(...), readDataFromApi(...), generateNewItem(...)]),
  asyncZip,
   ...
)
```

sithmel:
merge(apply)
should return a function taking an array
.bind([...])
should map the argument into "this" !?

I don't understand how this work.

But anyway, as I said a single function will come easier to remember.

conartist6:
The cool thing is that you could have `merge(call)` and `merge(apply)`, depending on whether the arguments needed to be spread or not.

I still don't like `multi`, because I don't find it descriptive. It could mean so many things, and I fear that it would come to have what amounted to more than one meaning. I wouldn't mind having `mergeApply` and `mergeCall` as syntactic sugar, since it would then be very clear what they were sugar for.

conartist6:
I edited the invocation of bind, as you noticed it was wrong.

conartist6:
An alternate formulation around `bind`: the curry helper, now in use everywhere, could defined a `partial` method on the function prototype. It would be the same method everywhere, it would basically do partial application of iterables, in case there was more than one iterable and you didn't wish to specify them all at the same time.

Examples:
```js
zip.partial(iterableA).partial(iterableB)()
// or
pipeExec(
  csvReader(...),
  asyncFork(3),
  mergeApply.partial([readDataFromMongo(...), readDataFromApi(...), generateNewItem(...)]),
  asyncZip,
)
```

sithmel:
merge(apply).bind(null, [readDataFromMongo(...), readDataFromApi(...), generateNewItem(...)]),

This implies that "merge" takes as a argument a variable number of iterables, instead of a single iterable of iterables.

I still think that would be a mistake. The api should be consistent within itself, to ease function composition. You have functions returning a iterable of iterables (fork, partition, split etc.) and therefore you need corresponding functions that take an iterable of iterables as argument.
But this is discussion for another ticket (likely #185 )

mergeApply tells that you are merging, and applying but doesn't describe how. And this risks developers to assume the 2 functions are combined in a way that might not be the one you intended.

Also as I said here https://github.com/sithmel/iter-tools/pull/167#issuecomment-464400728 the apply you propose spreads the arguments, and this doesn't.

conartist6:
I imagine it would work like call and apply.

`merge(call, fns, arg1s, arg2s, arg3s)`
`merge(apply, fns, [...args1], [...args2], [...args3])`

You say that `mergeApply` indicates that you are merging but doesn't describe how. Interestingly that's the criticism I have of `applyMulti`. It say that you are applying, but doesn't describe how.

I'm suggesting the reason that `mergeApply` tells you exactly what's going on is because the documentation can simply read: this function is syntactic sugar for `merge(apply)`. Now one of two things will be the case: either `mergeApply` will be showing up in a codebase where this pattern is so common that after looking it up once there will be no way for the coder to forget such a simple piece of documentation`, or alternatively it will be a codebase where this operation is not done often, and the smart coder will go to the minimal effort of writing out `merge(apply)`.

Honestly I think it would be easier to just not have the sugar method. It requires the programmer to import two things instead of one, and usage is three extra characters (a set of parens and a comma). Not a big deal.

sithmel:
I will review this. I'll be away, for a few days

conartist6:
Alright, no problem.

sithmel:
I had a second look at the comment here: https://github.com/sithmel/iter-tools/pull/167#issuecomment-464402849

I don't understand how these are supposed to work:
`merge(call, fns, arg1s, arg2s, arg3s)`
`merge(apply, fns, [[...args1], [...args2], [...args3]])`

> You say that `mergeApply` indicates that you are merging but doesn't describe how. Interestingly that's the criticism I have of `applyMulti`. It say that you are applying, but doesn't describe how.

That's why I prefer ```multi```: It is not creating expectations that might eventually be false.

conartist6:
And that's why I prefer the examples you see there. There's nothing hidden at all.
```js
import {merge, apply} from 'iter-tools';

merge(apply, [add, subtract, multiply], [[1, 2], [3, 4], [5, 6]]) // Iterable[3, -1, 30]
```

Let's break down how it works. `merge` takes a merging function and iterables. The function receives a value from each iterable as positional arguments, and it returns the value which should be emitted.

`apply` is being used as the merging function. At each iteration it receives one value from each of merge's iterables. It receives a function and an array. All that's left is for iter-tools to provide a function of the signature `apply(fn, argsArray)`.

`call` works similarly, except that it may provide more than two arguments to the merging function, which is fine because `call` will have the signature `call(fn, ...args)`.

sithmel:
Merge takes a function and an array of iterables.
https://github.com/sithmel/iter-tools/blob/1da1c3f9d558e940166ccf228f006ba37454f1ab/src/internal/async-merge-factory.mjs#L2

It should be like this:
```js
import {merge, apply} from 'iter-tools';

merge(apply, [[add, subtract, multiply], [[1, 2], [3, 4], [5, 6]]]) // Iterable[3, -1, 30]
```
right ?

conartist6:
No, `merge` would receive a variable number of iterables, as our other functions like `zip` do at present.

I think in the case of merge especially underlines the power of consistency here. It has a variable number of arguments to the `merge` call, and a variable number of arguments to the `merger` function. I think this use case shows perfectly why that amount and kind of consistency is so powerful.

conartist6:
Oh I see the code I wrote previously does do that, but in the mean time we settled on varargs, right?

conartist6:
To make one last clarification, I'm talking about the implementation of merge from https://github.com/sithmel/iter-tools/pull/174

sithmel:
@conartist6 I didn't bury the hatchet on the topic of multiple iterables vs iterable of iterables.

It seems clear to me that we are not going anywhere with this. Not until merge, collate and company are settled. I would suspend this PR and put forward some proposal for merge in an issue. I won't comment on #174 because I think the discussion diverged substantially from the implementation.

Issue #166, opened 2019-01-27
consumeMany, asyncConsumeMany


sithmel:
see https://github.com/sithmel/iter-tools/issues/151#issuecomment-457960146

sithmel:
closing as not necessary

PR #168, opened 2019-01-30
consume many (do not merge yet)

sithmel:
#166 
This is pretty much ready, but I found a bug in the asyncMerge that I want to fix before merging


sithmel:
This will be closed because of #174.
Before that I would like to fix the bug in asyncMerge so that I can use that code in another function.

Issue #169, opened 2019-01-30
Merge issues


conartist6:
We should have a lint rule to prevent this kind of breakage, but we don't. It is not safe in this library to use named exports in any file with a default export, because we use a babel plugin which transforms files with only default exports to `module.exports = exports.default`. When we export multiple functions from `merge.js`, this code is breaking.

sithmel:
As described above there are issues importing merge from node

sithmel:
I am writing here the outcome of the discussion on https://github.com/sithmel/iter-tools/pull/174

There will be different "merge" functions:
* collate/asyncCollate (merge by order)
* interleave/asyncInterleave (merge by position)
* asyncInterleaveReady (merge by readiness)

There will also be a low level implementation. See this PR as a reference: https://github.com/sithmel/iter-tools/pull/174

@conartist6 is working on a PR

conartist6:
This is fixed by the removal of `merge` in 7.x. The final advice on replacements to use is:
- `collate`/`asyncCollate` (merge by order or position)
- `asyncInterleaveReady` (merge by readiness)
- `interleave` (custom merges, including random ones)

PR #174, opened 2019-01-31
Collate and related proposals

conartist6:
See Changelog for description of changes


conartist6:
Looking for feedback before I finish it off with type tests and new method tests

sithmel:
I like some things, but I disagree with others:

### rename merge collate
I don't have strong opinions on that. They both make sense in my mind. The original naming was picked up from here https://github.com/bustle/streaming-iterables#merge

### removing merge helpers
They were problematic because of this: https://github.com/sithmel/iter-tools/issues/169
Probably they should have been in the cookbook.
But I would have rather done something different:
leaving the current function that gets all items and return an index:
```js
const minNumber = (items) =>
  items
    .map((item, index) => ({ index, item })) // add index
    .filter((decoratedItem) => !!decoratedItem.item) // remove null
    .sort((a, b) => a.item.value - b.item.value)[0].index // sort and pick first index

collate(minNumber, [iterable1, iterable2, iterable3])
```
for common cases use an object:
```js
collate({ strategy: 'byWeight', weight: [2, 2, 1] }, [iterable1, iterable2, iterable3])
```
Your version is simpler but less flexible.

### mergeByReadiness
This is missing but it is a very useful feature (it is also buggy, but this is another story). Maybe can be reintroduced as common strategy:
```js
asyncCollate({ strategy: 'byReadiness' }, [iterable1, iterable2, iterable3])
```
Not fully convinced of the name though

### merge/asyncMerge mergeAll/asyncMergeAll
I am against these. You can get exactly the same result using zip/zipAll and map.

sithmel:
## merge/zip
The 2 things are really very close only 3 character difference:
```js
map(func, zip(iterables))
// and
merge(func, iterables)
```
Despite being more complex I think a lot of people will be able to recognise map/zip without looking at the manual. "zip" and "map" are very common while "merge" is more ambiguous.

Performance wise, they should be very close. Just an extra function invocation. I am not sure performance alone justifies an extra (actually 4 extra) functions.

## collate
>I think the byReadiness helper could/should be a separate function.

That makes sense. (Also the current implementation of the callback is broken #168 )

I understand your point about the complexity of *merge* callback. Your solution is a reasonable compromise :+1:

conartist6:
I'm not thinking of the single extra function invocation, I'm thinking of all the arrays of items that are allocated by zip. A while ago i conceded that we should not try to optimize functions like entries and zip by reusing the arrays, however I do not concede that we should create core APIs that are likely to thrash the memory allocator. As a coder you almost always have a trade off to make between memory and CPU usage. Iterables represent a choice to trade extra cpu cycles for reduced memory usage. For user facing applications this is a far more favorable trade off: the users is only likely to be using one or two of them at a time, and modern processors are more than fast enough to blaze through whatever the application code is, so long as it is not needlessly adding complexity classes. On the other hand, many users have a crazy number of applications running at a time, and they expect each application to be responsive all the time -- nobody likes a page freezing up for even half a second while they're in the middle of their touch scrolling. So, to the extent that we can, I think that to serve the frontend constituency we should offer the APIs which can be implemented without any unnecessary memory allocation in a form that does not involve unnecessary memory allocation.

sithmel:
Have you tried running some benchmark ?

sithmel:
Also in theory you are passing an array to the merge function too.

conartist6:
I will look for evidence, but what is needed in this case isn't a benchmark but a profile. I don't care exactly how fast it is -- it is likely to be a little slower. I care that its other characteristics are favorable. I'm passing an array to the merge function, but it's the same array every time, something I can get away with only because I explode it into arguments onto the stack, precluding the possibility of the user mutating it.

sithmel:
It is important we have some way to profile memory consumption. I did that in long running tasks to debug memory leaks, I wonder if we can use the same technique.

In the mean time, I don't want to stop this PR. It provides a few extra functions, perhaps I'll find them useful at some point.

Let me know when this is ready to be merged. I'll review the implementation in the next days

conartist6:
It looks like we could use [v8-profiler](https://www.npmjs.com/package/v8-profiler) from node, then load the results into chrome, as described [here](https://github.com/node-inspector/v8-profiler/issues/125).

sithmel:
Crazy idea, curious to know what you think about it. It is about this problem:

>  [...] it becomes impossible to distinguish the iterable yielding undefined from the iterable being exhausted

What if we use a sparse array for that ?

That way you can easily find the minimum with:

```js
const minNumber = (items) =>
  items
    .map((value, index) => ({ index, value }))
    .sort((a, b) => a.value - b.value)[0].index

collate(minNumber, [iterable1, iterable2, iterable3])
```

This also makes possible to check if an iterable is exhausted:
```js
const arr = new Array(3)
arr[1] = 1
'0' in arr === false
'1' in arr === true
'2' in arr === false
```

Of course we can keep a couple of helpers for the common cases (byPosition and bySortingOrder)

conartist6:
I like a sparse array a lot better than exposing iterator items. It does unfortunately force the user to check membership in the sparse array even if they know their iterable won't return undefined, which I still don't think is great. I also think it still makes sense for the default interpretation of a function to be a comparator. Thought: if we do pass a sparse array, we could use arguments.length to figure out whether the passed function is a comparator or not. This is a bit evil, but we're pretty sure that comparators will always have `arguments.length === 2`, and functions expecting the sparse array would have `arguments.length === 1`. Alternatively we could just prefix an extra optional boolean called, say, `custom`, such that a custom collation would look like: `collate(true, sparseArr => { ... }, ...iterables)`

sithmel:
```js
// with custom function
collate((items) => ..., [iter1, iter2, iter3])

// use ordering
collate({ sorted: (a, b) => {} }, [iter1, iter2, iter3])

// use position
collate({ byPosition: 0 }, [iter1, iter2, iter3])
```
Or something similar ?

conartist6:
What if we did this?
```js
higherOrderCombine(function * (buffer1, buffer2) {
   while (true) {
     yield buffer1.isEmpty() ? buffer1.take() : buffer2.take()
   }
}, iterable1, iterable2)
```

`take()` would be implemented as `this.shift(); this.push(source.next)`.

It has several advantages. It's very flexible. It allows easier expression of more complicated strategies, and even intermixing other items which aren't part of any source. It unambiguously signals when a source is exhausted. `shift` could either return undefined or throw an error when the buffer has no items -- I'm not sure which would be better, but I think either would be OK. It would call return on the generator when all the buffers were empty.

sithmel:
Not bad, but it looks to me a new interface to learn. I would feel more at home picking up one item on every iteration. Also I don't know how to wire this new abstraction with the return

conartist6:
I think I know how to wire return.

Could you explain what you mean by "a new interface to learn"? It seems to me that that's not necessarily bad, particularly if our API is then better able to mirror the structure of the problem it is made to solve. In addition other methods might benefit from a similar interface as well. `asyncBuffer` comes immediately to mind, and could then have `buffer` alongside its async counterpart, a function which could be used to implement this one, and whose documentation would also help users understand what was going on here.

And I think that pick function code written with this interface will be more readable. It should be largely self-evident what it does even to a reader who has not first consulted the documentation, in no small part because the generator itself contains the `yield` statement which unambiguously defines the most important part of the behavior: what values can be expected as a result.

And much more so than sparse arrays and returning numbers, it makes explicit the most complicated thing about writing these functions: that you are forced to deal with what happens when one iterable is exhausted. If the programmer makes a mistake and tries to yield from an exhausted iterable, we can even throw an error whose stack trace will point to the exact like of code where the problem is!

Considering the alternative interface for a second, the sparse array of items, you can't destructure it without losing track of the sparseness, so while the iterables are almost certain to have names which are likely relevant to the collation you'll be forced to refer to them by index in order to be able to do the `has` check. Essentially a rather simple use case ends up looking like:

```js
collate(sparseValues => {
  return sparseValues.findIndex((item, i) => {
    if (!(i in sparseValues)) return true
    // here, where the logic is written, the name of item is lost
  }
}, a, b)
```

sithmel:
> Could you explain what you mean by "a new interface to learn"? 

I think you understood, be cause you pretty much answered to me

> Considering the alternative interface for a second, the sparse array of items, ....

you don't have to convince me that is not perfect. It has its shortcomings

I found a couple of examples where basically they avoided the problem creating different functions for the most common cases (round robin, the fastest one, by order):

https://github.com/bustle/streaming-iterables#merge
https://more-itertools.readthedocs.io/en/stable/api.html#combining

Should we do it that way eventually ?

conartist6:
Well the fastest one is always going to need its own method. I don't have a particular opinion on the difference between a `roundRobin` method and an overload of collate that takes a number parameter. If we implemented a combine higher order generator we could use it to create the implementations of `collate(comparator)`, `roundRobin`, and `earlyRobin` or whatever.

sithmel:
Makes sense

conartist6:
So I like what more-itertools has done by having `collate` specifically be designed to work with sorted inputs. Then `interleave` does the same thing essentially for unsorted inputs, which is the same as what it does in Immutable.

From their docs on interleave_longest:
> This function produces the same output as roundrobin(), but may perform better for some inputs (in particular when the number of iterables is large).
Uh, what???

conartist6:
Just rebased this. I had to make significant changes to curry, which was previously incompatible with variadic functions which also take config arguments.

conartist6:
Those changes should be separated out into a new commit too, which can be done, though it would be good to know that you approve of them first.

conartist6:
@sithmel @KSXGitHub This branch now embodies my proposal as to the way forward for pipelining, and particularly the operation formerly known as `mapFunc`.

Instead of `mapFunc`, a higher order call would look like:
```js
import { fork, merge, call, zip, pipeExec } from 'iter-tools'
const add = (a, b) => a + b
const multiply = (a, b) => a * b
pipeExec(
  fork(2, [3, 3]),
  merge(call).partial([add, multiply]),
  zip
) // [6, 9]
```

conartist6:
I have added a test to `pipeline.test.js` that demonstrates this technique in action.

src/internal/curry.mjs
```diff
@@ -1,10 +1,20 @@
...
     if (expectedArgsLength <= args.length) {
       return fn(...appliedArgs, ...args)
     }
-    return curry(fn, expectedArgsLength - args.length, [...appliedArgs, ...args])
+    const argsLength = args.length
+    args.unshift(...appliedArgs)
+    return curry(fn, expectedArgsLength - argsLength, args)
   }
+
+  curryInternal.partial = function (...args) {
```
KSXGitHub: Dot pattern looks pretty, but it may prevent tree-shaking. Why don't you use `partialCurryInternal` instead?

conartist6: You are in fact both right and wrong. The . pattern does not prevent tree shaking. You can see that clearly [here](https://rollupjs.org/repl?version=1.2.3&shareable=JTdCJTIybW9kdWxlcyUyMiUzQSU1QiU3QiUyMm5hbWUlMjIlM0ElMjJtYWluLmpzJTIyJTJDJTIyY29kZSUyMiUzQSUyMmltcG9ydCUyMCU3QnNxdWFyZSUyQyUyMGN1YmUlN0QlMjBmcm9tJTIwJy4lMkZtYXRocy5qcyclNUNuJTVDbmNvbnNvbGUubG9nKHNxdWFyZSg1KSklM0IlNUNuJTJGJTJGY29uc29sZS5sb2coY3ViZSg1KSklM0IlMjIlMkMlMjJpc0VudHJ5JTIyJTNBdHJ1ZSU3RCUyQyU3QiUyMm5hbWUlMjIlM0ElMjJtYXRocy5qcyUyMiUyQyUyMmNvZGUlMjIlM0ElMjJpbXBvcnQlMjBzcXVhcmUlMjBmcm9tJTIwJy4lMkZzcXVhcmUuanMnJTNCJTVDbiU1Q24lMjBmdW5jdGlvbiUyMGN1YmUlMjAoeCklMjAlN0IlNUNuJTVDdHJldHVybiUyMHNxdWFyZSh4KSUyMColMjB4JTNCJTVDbiU3RCU1Q24lNUNuY3ViZS5mb28lMjAlM0QlMjBmdW5jdGlvbigpJTIwJTdCJTdEJTVDbiU1Q24lMjBleHBvcnQlMjAlN0IlMjBjdWJlJTJDJTIwc3F1YXJlJTIwJTdEJTIyJTJDJTIyaXNFbnRyeSUyMiUzQWZhbHNlJTdEJTJDJTdCJTIybmFtZSUyMiUzQSUyMnNxdWFyZS5qcyUyMiUyQyUyMmNvZGUlMjIlM0ElMjJleHBvcnQlMjBkZWZhdWx0JTIwZnVuY3Rpb24lMjBzcXVhcmUlMjAoJTIweCUyMCklMjAlN0IlNUNuJTVDdHJldHVybiUyMHglMjAqJTIweCUzQiU1Q24lN0QlMjIlMkMlMjJpc0VudHJ5JTIyJTNBZmFsc2UlN0QlNUQlMkMlMjJvcHRpb25zJTIyJTNBJTdCJTIyZm9ybWF0JTIyJTNBJTIyY2pzJTIyJTJDJTIybmFtZSUyMiUzQSUyMm15QnVuZGxlJTIyJTJDJTIyYW1kJTIyJTNBJTdCJTIyaWQlMjIlM0ElMjIlMjIlN0QlMkMlMjJnbG9iYWxzJTIyJTNBJTdCJTdEJTdEJTJDJTIyZXhhbXBsZSUyMiUzQW51bGwlN0Q=)

But also, the curry function itself does very much prevent all tree shaking. We need to go through and mark every call site of `iterableCurry` and `asyncIterableCurry` with `#__PURE__ ` comments.

conartist6:
@KSXGitHub one big head scratcher is that I don't know how to type the partial function in typescript. I sort of doubt that reasonable types are even possible, even in flow.

src/internal/curry.mjs
conartist6: Also adding the pure comments will ensure that it doesn't matter what happens inside curry.

KSXGitHub:
@conartist6 

> one big head scratcher is that I don't know how to type the partial function in typescript. I sort of doubt that reasonable types are even possible, even in flow.

It is possible (since TypeScript 3.0+), but before I show you the type definition, I'd like to know if this is even necessary (`curry` and `partial` are used internally only, right?).

conartist6:
`curry` is only used internally, but `partial` is intended for external consumption.

KSXGitHub:
> `curry` is only used internally, but `partial` is intended for external consumption.

Then it is even easier to make type definitions.

`curry` is only used internally, so we don't need a generic `curry`/`partial`.

But we still need to definite `partial` in every curried function, for example:

```typescript
declare function map<A, B> (fn: (x: A) => B): {
  (iterable: Iterable<A>): Iterable<B>
  partial: /* partial type here */
}
```

Now I want to know some real use case of `partial` method.

src/internal/__tests__/curry.test.js
```diff
@@ -41,7 +41,52 @@ describe('variadicCurryWithValidation', function () {
     const f = variadicCurryWithValidation((x) => x === null, 'null', (x) => 0, func, false)
     expect(() => f(1)(2)(3)).toThrowError(new Error('func takes up to 2 arguments, followed by null. You already passed 3 arguments and the last argument was not null'))
   })
-  it('works on functions which take a variable number of iterables', function () {
 
+  describe('explicit partial application', function () {
+    let curriedMethod
+
+    beforeEach(function () {
```
KSXGitHub: Instead of `let`/`beforeEach`, you can use a factory function (function that returns `curriedMethod`)

```javascript
const makeCurriedMethod = variadicCurryWithValidation(/* ... */)

it('curried methods have a partial method', function () {
  expect(typeof makeCurriedMethod()).toHaveProperty(expect.any(Function))
})
```

conartist6:
Added an implementation of the function I'm calling `interleaveGenerator` which I propose should replace the no-helper merge. Instead of a callback which returns an item index, `interleaveGenerator` is a higher order function over a generator, which sets up certain inputs to the generator function.

As to the name, I'm not 100% satisfied with it. It sounds like it interleaves generators, when what I want to convey is that it creates an interleaving generator function from a normal generator function. In particular, I think it is sensible to advertise this to be a decorator function. It serves the purpose of a decorator, and actually once TC39 finds a way to allow decorators on normal functions it can be used effectively with a decorator syntax, like so:
```js
@interleaveGenerator(a, b)
function my interleave(isDone, a, b) {
  while (!isDone()) {
    if (!a.exhausted()) yield a.take()
    if (!b.exhausted()) yield a.take()
  }
}
```

conartist6:
I have no desire to make explicit partial types for every type. That would be a nightmare. In all likelihood this feature would be more useful to programmers writing untyped code.

The use case is as documented in [my previous comment](https://github.com/sithmel/iter-tools/pull/174#issuecomment-466710061). Partial is necessary here as opposed to normal currying, because with normal currying there is no way to specify some but not all input iterables.

Since the case that is most important for `partial` is functions taking `...iterable`, this suggests that we could perhaps add a limited type definition just for the variadic functions that need it most: `merge`, `mergeAll`, `concat`, `zip`, and `zipAll`.

src/internal/__tests__/curry.test.js
```diff
@@ -41,7 +41,52 @@ describe('variadicCurryWithValidation', function () {
...
+
+    it('can be used to defer application of the data parameter', function () {
+      expect(curriedMethod.partial().partial('hello ').partial('world').partial('🙂')()).toBe('hello world! 🙂')
+    })
+  })
+
+  describe('currying variadic functions', function () {
+    let sumMethod
+
+    beforeEach(function () {
```
KSXGitHub: [This too](https://github.com/sithmel/iter-tools/pull/174/commits/a70990e282568420cfc70bbcfc98992e36974a7a#r259602797).

You are reassigning one `sumMethod` variable each time a new test run, this would be a problem when the tests are run in parallel.

conartist6: Jest never parallelizes inside a file. It will run different test files in parallel, and different projects as well, but each file is given it's own fresh execution environment and the tests inside are always run synchronously. Even if the tests are async, inside a single file each will wait for the previous to finish.

src/internal/curry.mjs
KSXGitHub: In [your example][conartist example], `cube.foo` is defined in global context, thus tree-shakable. But your code is more akin to [this][my example] (every method is defined in local context, just like `partial`).


[conartist example]: https://rollupjs.org/repl?version=1.2.3&shareable=JTdCJTIybW9kdWxlcyUyMiUzQSU1QiU3QiUyMm5hbWUlMjIlM0ElMjJtYWluLmpzJTIyJTJDJTIyY29kZSUyMiUzQSUyMmltcG9ydCUyMCU3QnNxdWFyZSUyQyUyMGN1YmUlN0QlMjBmcm9tJTIwJy4lMkZtYXRocy5qcyclNUNuJTVDbmNvbnNvbGUubG9nKHNxdWFyZSg1KSklM0IlNUNuJTJGJTJGY29uc29sZS5sb2coY3ViZSg1KSklM0IlMjIlMkMlMjJpc0VudHJ5JTIyJTNBdHJ1ZSU3RCUyQyU3QiUyMm5hbWUlMjIlM0ElMjJtYXRocy5qcyUyMiUyQyUyMmNvZGUlMjIlM0ElMjJpbXBvcnQlMjBzcXVhcmUlMjBmcm9tJTIwJy4lMkZzcXVhcmUuanMnJTNCJTVDbiU1Q24lMjBmdW5jdGlvbiUyMGN1YmUlMjAoeCklMjAlN0IlNUNuJTVDdHJldHVybiUyMHNxdWFyZSh4KSUyMColMjB4JTNCJTVDbiU3RCU1Q24lNUNuY3ViZS5mb28lMjAlM0QlMjBmdW5jdGlvbigpJTIwJTdCJTdEJTVDbiU1Q24lMjBleHBvcnQlMjAlN0IlMjBjdWJlJTJDJTIwc3F1YXJlJTIwJTdEJTIyJTJDJTIyaXNFbnRyeSUyMiUzQWZhbHNlJTdEJTJDJTdCJTIybmFtZSUyMiUzQSUyMnNxdWFyZS5qcyUyMiUyQyUyMmNvZGUlMjIlM0ElMjJleHBvcnQlMjBkZWZhdWx0JTIwZnVuY3Rpb24lMjBzcXVhcmUlMjAoJTIweCUyMCklMjAlN0IlNUNuJTVDdHJldHVybiUyMHglMjAqJTIweCUzQiU1Q24lN0QlMjIlMkMlMjJpc0VudHJ5JTIyJTNBZmFsc2UlN0QlNUQlMkMlMjJvcHRpb25zJTIyJTNBJTdCJTIyZm9ybWF0JTIyJTNBJTIyY2pzJTIyJTJDJTIybmFtZSUyMiUzQSUyMm15QnVuZGxlJTIyJTJDJTIyYW1kJTIyJTNBJTdCJTIyaWQlMjIlM0ElMjIlMjIlN0QlMkMlMjJnbG9iYWxzJTIyJTNBJTdCJTdEJTdEJTJDJTIyZXhhbXBsZSUyMiUzQW51bGwlN0Q=

[my example]: https://rollupjs.org/repl?version=1.2.3&shareable=JTdCJTIybW9kdWxlcyUyMiUzQSU1QiU3QiUyMm5hbWUlMjIlM0ElMjJtYWluLmpzJTIyJTJDJTIyY29kZSUyMiUzQSUyMmltcG9ydCUyMGNyZWF0ZUZ1bmN0aW9ucyUyMGZyb20lMjAnLiUyRmZhY3RvcnkuanMnJTVDbiU1Q25jb25zdCUyMCU3QiUyMGZvbyUyMCU3RCUyMCUzRCUyMGNyZWF0ZUZ1bmN0aW9ucygpJTVDbiU1Q25jb25zb2xlLmxvZyhmb28oKSklMjIlMkMlMjJpc0VudHJ5JTIyJTNBdHJ1ZSU3RCUyQyU3QiUyMm5hbWUlMjIlM0ElMjJmYWN0b3J5LmpzJTIyJTJDJTIyY29kZSUyMiUzQSUyMmV4cG9ydCUyMGRlZmF1bHQlMjBmdW5jdGlvbiUyMGNyZWF0ZUZ1bmN0aW9ucyUyMCgpJTIwJTdCJTVDbiU1Q3Rjb25zdCUyMHJlc3VsdCUyMCUzRCUyMCU3QiU3RCU1Q24lNUN0cmVzdWx0LmZvbyUyMCUzRCUyMCgpJTIwJTNEJTNFJTIwJ2ZvbyclNUNuJTVDdHJlc3VsdC5iYXIlMjAlM0QlMjAoKSUyMCUzRCUzRSUyMCdiYXInJTVDbiU1Q3RyZXN1bHQuYmF6JTIwJTNEJTIwKCklMjAlM0QlM0UlMjAnYmF6JyU1Q24lNUN0cmV0dXJuJTIwcmVzdWx0JTVDbiU3RCUyMiU3RCU1RCUyQyUyMm9wdGlvbnMlMjIlM0ElN0IlMjJmb3JtYXQlMjIlM0ElMjJjanMlMjIlMkMlMjJuYW1lJTIyJTNBJTIybXlCdW5kbGUlMjIlMkMlMjJhbWQlMjIlM0ElN0IlMjJpZCUyMiUzQSUyMiUyMiU3RCUyQyUyMmdsb2JhbHMlMjIlM0ElN0IlN0QlN0QlMkMlMjJleGFtcGxlJTIyJTNBbnVsbCU3RA==

src/internal/__tests__/curry.test.js
KSXGitHub: Concurrency is not the only reason to prefer [my style](https://github.com/sithmel/iter-tools/pull/174/commits/a70990e282568420cfc70bbcfc98992e36974a7a#r259602797), there're also several others:
  * More concise and straightforward
  * Allow VSCode's type hint to help
  * Should function creation throw an error, the tests would fail immediately (I am too lazy to check whether errors in `beforeEach` prevents tests from being run)

sithmel:
## remove zip, ziplongest in favour of zip/zipAll
**Ok**

## merge/mergeAll:

I am not sure about the implementation, I would rather review it in another ticket/PR.

## collate/asyncCollate:

The way it works I would rather have different functions dealing with most common cases. For example:

* collate/asyncCollate (merge by order)
* interleave/asyncInterleave (merge by position)
* asyncMergeByReadiness (merge by readiness)

## apply/call
I don't see other use cases for these, apart than making what I could do with mapFunc/applyMulti much more complicated.

## function signature for multiple iterables
I am not giving up on this. The api should be consistent within itself, to ease function composition. You have functions returning a iterable of iterables (fork, partition, split etc.) and therefore you need corresponding functions that take an iterable of iterables as argument.

## partial method
Again, it looks to me we are going to great lengths to avoid doing the easiest thing.



My general feeling about this PR is that it contains too many controversial changes. And it is simply impossible to discuss all of them in one place.

I propose to close it and restart from the low level merge.

conartist6:
I'm happy to split round robin collation out into interleave. I think I meant to do that but I forgot.

I also don't really plan to merge this PR anymore. It is here to express my vision as to the way we can offer a number of low level features that can work harmoniously to allow users to accomplish high level goals. As we come to consensus as to whether or not we want the various things in it I'll split them out as smaller pieces so that they can be more fully reviewed in isolation.

You've just called out one of the uses of `apply` yourself: it's highly useful when you have an iterable of arguments and a variadic function. I'd call it a new iterable sink, like `toArray` or `consume`.

I think TC39 has spoken on variadic functions. By offering call spread and arguments spread they embraced the pattern, and it makes sense too. Most of the time users of these variadic functions will have separate expressions to pass to each argument. In the less common case that they have an iterable, all they have to do is write `...`. In the even less common case that we are in a pipeline and working exclusively with higher order functions, we now have `apply`.

The variadic arguments arrangement has many desirable properties or the common case. For one, no extra syntax is needed. Another: it is also the most optimized, as it passes the variables on the stack instead of having to create an Array.

Finally Typescript endorses the pattern. Typescript has an awful time with tuples, and has a tendency to lose their types unless two things happen: 1) the tuple-typed array is defined in exactly the location that it is consumed by a type which supports tuple-typed arrays. and 2) that API has an overload for Array and an overload for Iterable, since there is no concept of iterable tuples. This means that in order for the common case type information to not be lost, the user has to make a call of the structure: `variadicFn(arg, [iterableA, iterableB])`. The `[]` cannot go anywhere else, and thus is just dead weight, we may as well have allowed the user to call the function in a variadic manner.

conartist6:
I'm thinking about how well `interleaveGenerator` works with async by imaging using it to write `asyncInterleaveReady`. Main adaptations: offering a `ready` promise on the buffer (which returns the buffer) and swapping the `isDone` method with one returning the inverse, called `canTake`.

```js
asyncInterleaveGenerator(async function * (canTake, ...buffers) {
  let buffer
  while ((buffer = await Promise.race(concat(canTake(), map(buffer => buffer.ready())))) && buffer) {
    yield await buffer.take()
  }
})
```

conartist6:
OK so finding a way to facilitate the more powerful usages that were possible with the current `merge` implementation was the last major roadblock to the merge/collate changes being ready (from my perspective). I'm happy with it now; after much tweaking I don't think I can design any better API. I'm going to start making the incremental PRs we can actually ship.

sithmel:
I am really intrigued by interleaveGenerator. I'd like to understand better how it works.

src/internal/__tests__/curry.test.js
conartist6: I do like it, actually. `beforeEach` will not run the test if it errors out, and it does work quite well when the setup can be expressed in a purely additive manner with nested describe blocks each adding a bit to the setup.

In simpler situations like this though it can be more useful to be able to pass a parameter to your setup, also there's the type hint, though I don't use vscode.

KSXGitHub: Reading your response, I can't know for sure if you agree or disagree with me, so I still insist on this (but I guess it didn't matter since you're probably not going to merge this).

BTW, since you don't use VSCode, what editor do you use? (I'm just curious)

conartist6: I'm a Sublime text user.

PR #175, opened 2019-01-31
more readable/static curry

conartist6:


src/internal/curry.mjs
```diff
@@ -1,11 +1,8 @@
-export default function curry (fn) {
-  function nest (len, args) {
-    return (...xs) => {
-      if (len - xs.length <= 0) {
-        return fn(...args, ...xs)
-      }
-      return nest(len - xs.length, [...args, ...xs])
+export default function curry (fn, expectedArgsLength = fn.length, appliedArgs = []) {
+  return (...args) => {
+    if (expectedArgsLength - args.length <= 0) {
```
KSXGitHub: Why not just `expectedArgsLength <= args.length`?

sithmel:
looks good to me

src/internal/curry.mjs
sithmel: I'll apply the tweak after the merge

PR #178, opened 2019-02-01
asyncFilter and asyncMap now work in all circumstances, and are more efficient 

conartist6:
Also: CircularBuffer can now contain a variable number of items. Application of new technique to async map is coming.

### DIFF
--- a/src/internal/async-map-parallel.mjs
+++ b/src/internal/async-map-parallel.mjs
@@ -1,6 +1,5 @@
-import { asyncIterableCurry, asyncify } from './internal/async-iterable'
-
-import CircularBuffer from './internal/circular-buffer'
+import CircularBuffer from './circular-buffer'
+import { asyncify } from './async-iterable'
 
 class ParallelMapper {
   constructor (iterable, mapFn, concurrency) {
@@ -76,18 +75,6 @@ class ParallelMapper {
   [Symbol.asyncIterator] () { return this }
 }
 
-async function * asyncMap (concurrency, func, iterable) {
-  if (concurrency == null) {
-    concurrency = 1
-  }
-
-  let c = 0
-
-  const runner = new ParallelMapper(iterable, async item => func(item, c++), concurrency)
-
-  for await (const value of runner) {
-    yield value.mappedValue
-  }
+export default function asyncMapParallel (iterable, mapFn, concurrency) {
+  return new ParallelMapper(iterable, mapFn, concurrency)
 }
-
-export default asyncIterableCurry(asyncMap, 2, 3)
--- a/jest.config.js
+++ b/jest.config.js
@@ -10,19 +10,22 @@ function makeProject (projectConfig) {
   }, projectConfig)
 }
 
-module.exports = Object.assign({
+module.exports = {
   testEnvironment: 'node',
   testMatch: [],
   coverageReporters: ['json-summary', 'text', 'lcov'],
   projects: [
-    // makeProject({
-    //   name: 'es5'
-    // }),
-    // makeProject({
-    //   name: 'es2018'
-    // })
-  ]
-},
-makeProject({
-  name: 'es2015'
-}))
+    makeProject({
+      name: 'es5'
+    }),
+    makeProject({
+      name: 'es2015'
+    }),
+  ].concat(
+    process.env.CI
+      ? makeProject({
+          name: 'es2018'
+        })
+      : []
+  )
+}
--- a/src/__tests__/cursor.test.js
+++ b/src/__tests__/cursor.test.js
@@ -1,10 +1,10 @@
 /* eslint-env node, jest */
-const { cursor, asyncCursor, asyncToArray } = require('..')
+const { cursor, map, asyncCursor, asyncToArray, asyncMap } = require('..')
 
 describe('cursor', function () {
   it('frames iterable', function () {
     const iter = cursor({ size: 3 }, [1, 2, 3, 4, 5])
-    expect(Array.from(iter)).toEqual([
+    expect(Array.from(map(wndw => [...wndw], iter))).toEqual([
       [undefined, undefined, 1],
       [undefined, 1, 2],
       [1, 2, 3],
@@ -15,7 +15,7 @@ describe('cursor', function () {
 
   it('frames iterable (use filler)', function () {
     const iter = cursor({ size: 3, filler: 'x' }, [1, 2, 3, 4, 5])
-    expect(Array.from(iter)).toEqual([
+    expect(Array.from(map(wndw => [...wndw], iter))).toEqual([
       ['x', 'x', 1],
       ['x', 1, 2],
       [1, 2, 3],
@@ -26,7 +26,7 @@ describe('cursor', function () {
 
   it('frames iterable (cursor equal to the sequence)', function () {
     const iter = cursor({ size: 5 }, [1, 2, 3, 4, 5])
-    expect(Array.from(iter)).toEqual([
+    expect(Array.from(map(wndw => [...wndw], iter))).toEqual([
       [undefined, undefined, undefined, undefined, 1],
       [undefined, undefined, undefined, 1, 2],
       [undefined, undefined, 1, 2, 3],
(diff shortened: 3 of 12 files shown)


conartist6:
OK this is where it gets really complicated. There's a bug in what I have here, which is that sequencing is lost. If one mapper returns earlier, it can emit its item out of order.

conartist6:
Oh ok scratch what I said

conartist6:
I'm going to deal with that tomorrow though. Staying up til 2 last night working on this was bad idea >_<

I also think there is also likely to be enough overhead involved in all this that it will make sense to special case concurrency one to avoid all this logic

src/internal/async-map-parallel.mjs
```diff
@@ -0,0 +1,52 @@
...
+
+        if (!buffer.isFull()) {
+          // Go on filling up the buffer in the background
+          this.next()
+        }
+
+        return parallelPromise
+      })
+
+      buffer.push(promise)
```
conartist6: This is a problem. It needs to be combined with line 35, or else we return an item *and* buffer it which doesn't make sense. However something else is wrong, because when I do that other code breaks.

package.json
```diff
@@ -55,6 +55,7 @@
     "@babel/runtime-corejs2": "^7.1.5",
     "dequeue": "^1.0.5",
     "little-ds-toolkit": "^1.1.0",
+    "p-defer": "^1.0.0",
```
sithmel: Modules from sindresohrus are never transpiled to es5. This will break the es5 version

sithmel:
I'd like to understand better how the new buffer and async-map-parallel work. I am sure everything will be clearer with some extra test.

I have so underestimated the complexity of this! :man_facepalming: 

Thanks for you efforts.

package.json
KSXGitHub: I think should just drop es5 and legacy supports.

conartist6:
This should now be pretty much ready to go.

conartist6:
Yeah I had also underestimated how complex it would be, but I don't see any simpler way, and I've spent quite a bit of time thinking about it. The good news is that that means it's a really useful thing for iter-tools to provide. Easy to describe, but clearly quite tricky to implement.

conartist6:
Added some in-depth comments.

Also I forgot to add earlier that there's a bit of a potentially breaking change in cursor: since you're now getting the same cursor instance at every iteration, if you should manage to hold on to a cursor between iterations you must now create a copy of it (with `[...cursor]`) by hand.

conartist6:
The one outstanding issue is that this PR needs what async-buffer purports to offer. I think I need to either make use of the existing implementation or if I can't, I need to figure out what's wrong with it and how to make it usable

conartist6:
I'm not sure if we want to do this, but I think the change that would need to happen is that the async-buffer api would change: instead of it making a buffer for you, you would pass it a buffer. I like that, because as I've always said I don't see that function as being very useful if you can't get at the buffer, which I believe is why I couldn't use it in this case when I needed what it purports to provide. If you provide it the buffer you definitely have access to it yourself, and in this case it would allow me to pass a subclass of buffer with an overridden push method, like so:
```js
class MappingBuffer extends CircularBuffer {
  push(sourceItem) {
    super.push(sourceItem.then(item => {
       // initiate parallel promise
    })
  }
}

asyncBuffer(new MappingBuffer(), iterable)
```

The thing is, we'd have to export our circular buffer class if we want to give users that power as well. I don't think it's such a bad idea  -- as we've found it's frequently a pretty useful construct, it is in fact an iterable, and if some of my other changes are accepted we'll be exposing instances of it directly to users anyway.

conartist6:
Let's assume that if that if we do change async-buffer we will do so in a separate PR. I'm ready to merge this.

sithmel:
@conartist6 let me review this before merging. Apologies if it is taking me a bit of time :-)

conartist6:
Realized I was only running the es2015 tests. Oversight when merging the jest diff. Fixed it here. I've configured jest to only run the es2018 tests in CI, where we are sure that node is v10. I anticipate es2018 only bugs will be pretty rare and we still will prevent them getting into our codebase without restricting development to only those with the newest node version.

conartist6:
And no worries. I understand that it's pretty tricky to review.

src/internal/async-map-parallel.mjs
```diff
@@ -0,0 +1,80 @@
+import CircularBuffer from './circular-buffer'
+import asyncify from './asyncify'
+
+class ParallelMapper {
+  constructor (iterable, mapFn, concurrency) {
+    this.iter = asyncify(iterable)[Symbol.asyncIterator]()
```
sithmel: why do you use asyncify here and not ensureAsyncIterable ?

src/internal/async-map-parallel.mjs
```diff
@@ -0,0 +1,80 @@
...
+    this.buffer = new CircularBuffer(concurrency)
+    this.mapFn = mapFn
+    this.done = false
+    this.previousUnresolvedPromise = null
+  }
+
+  next () {
+    const { iter, buffer, done, mapFn } = this
```
sithmel: I can't see how you manage to call "return" on iter

src/internal/async-map-parallel.mjs
```diff
@@ -0,0 +1,80 @@
...
+              // to wait until we had produced one mapped item until our consumer could call our
+              // next method again triggering another source fetch. This would result in completely
+              // serial behavior.
+              buffer.push(this.next())
+            }
+          }
+
+          let mappedValue
+          if (!item.done) {
+            mappedValue = await mapPromise
```
sithmel: you are awaiting the return of of the mapFn. This was not how it was working before.
The point of asyncMap - asyncFilter was to running the these functions in parallel

sithmel: Never mind, I think I understand now. Giving that you are inside a "then" callback, this will happen in parallel

src/internal/async-map-parallel.mjs
```diff
@@ -0,0 +1,80 @@
...
+    this.done = false
+    this.previousUnresolvedPromise = null
+  }
+
+  next () {
+    const { iter, buffer, done, mapFn } = this
+    let promise
+
+    if (buffer.isFull() || done) {
+      promise = buffer.shift()
```
sithmel: I find it easier to read if you return promise at the end of this block

src/internal/async-map-parallel.mjs
```diff
@@ -0,0 +1,80 @@
...
+
+      if (!done) {
+        // Keep the buffer full!
+        buffer.push(this.next())
+      }
+    } else {
+      promise = iter.next()
```
sithmel: Same as above:
```return iter.next() ...```

sithmel:
About the proposed changes to async-buffer. Let's keep them outside this thread. We can discuss them in another ticket

conartist6:
Super bowl time. Obligations of being an american you know. Normal coding
operations will return later

sithmel:
I am available to help on this @conartist6. Is that ok if I create a new branch starting with this, and I propose some changes on that ?

sithmel:
In the mean time I merged master into this branch

conartist6:
I wish you would not do that. I'll undo it for now, but it really messes up my workflow when you merge branches into mine. I will rebase my work on master.

conartist6:
OK you were right of course that asyncFilter should just call asyncMap. The diff is updated.

sithmel:
> I wish you would not do that. I'll undo it for now, but it really messes up my workflow when you merge branches into mine. I will rebase my work on master.

Sorry, I just wanted to be helpful :-D

@conartist6 
In general I am quite happy about this. I have just a few minor things, that I'd like to change.
I can:
* merge this and prepare a new PR with some changes
* create a branch out of this with the changes
* pointing to the changes here in the comments and let you implement them

conartist6:
I'm content for the branch to be merged, even if more changes are planned. Why don't you go ahead and do that, but out of curiosity what are you thinking?

sithmel:
> out of curiosity what are you thinking?

Really minor things, you will see soon :-)

Issue #180, opened 2019-02-03
asyncFork() does not work (master branch)


KSXGitHub:
```
TypeError: (0 , _ensureAsyncIterable.default)(...)[Symbol.iterator] is not a function
  at asyncFork (es2015/async-fork.js:31:20
```

sithmel:
How can I reproduce the problem ? my tests are passing

KSXGitHub:
> my tests are passing

Your tests do not execute `asyncFork`, only `fork`.

https://github.com/sithmel/iter-tools/blob/f9a264e7f02448b3bcb29b3dd20c67b3b9e5e7f7/src/__test__/fork.test.js#L2

sithmel:
:facepalm:

KSXGitHub:
**Suggestion:** Add `--coverage` to your `jest` command to display coverage.

sithmel:
great advice @KSXGitHub 
I noticed that it makes tests super slow.
I am going to add a test:coverage script :+1:

sithmel:
added tests and fixed master

coverage command was always there under my nose ```npm run coverage```

PR #181, opened 2019-02-03
Create splitAt() and asyncSplitAt()

KSXGitHub:
Fixes https://github.com/sithmel/iter-tools/issues/134

**TODOs:**
  - [ ] `asyncSplitAt`: use `asyncFork` in place of `asyncTee` (wait for https://github.com/sithmel/iter-tools/issues/180)
  - [x] Improve TypeScript definitions
  - [ ] README and CHANGELOG


sithmel:
> I would like to see a better implementation. In particular I want to see that if the user consumes the entire first half of the split prior to consuming any of the second, that no cache values are created.
If the user begins consuming the second half of the split prior to the first, I want to see and array cache of the correct size created (new Array(position)) and filled by calling next() position times, which must be done by hand in order to avoid closing the source iterable accidentally.
This means that you will also need a try..finally block to ensure that iter.return() is called if splitAt returns prior to exhausting iter.

This makes sense. The optimised implementation for this, is going to be way more complex. But also super interesting!

KSXGitHub:
@conartist6 You seem to be demanding optimization, but I want to avoid premature optimization, so unless this is absolutely necessary, I prefer simple implementation with small size of code.

KSXGitHub:
> I am happy to work on it if you want @KSXGitHub ;-)

I would be happy if you take care of this. However, I still prefer smaller size, optimization is OK as long as it does not require twice as much code.

sithmel:
The point of using iterables is being memory friendly, as much as it is possible.

Arrays beat iterables every single time in benchmarks. It would be pointless having a memory friendly alternative that is not optimised for memory usage.

I spent my days reducing bundle sizes, and I am not happy to add unnecessary bloat. But trade offs are necessary.

conartist6:
```js
import ensureIterable from './internal/ensure-iterable'

export default function splitAt(index, iterable) {
  const iter = ensureIterable(iterable)[Symbol.iterator]()
  let position = 0
  let firstCache = null
  let firstCacheStart = null
  let done = false

  function cleanUp() {
    if (!done && typeof iter.return === 'function') {
      iter.return()
    }
  }

  function * first() {
    try {
      while (!done) {
        if (firstCache) {
          yield firstCache[position - firstCacheStart]
        } else {
          const item = iter.next()
          done = item.done
          if (!done) {
            yield item.value
          }
        }
        position++
      }
    } finally {
      if (position !== index - 1) {
        cleanUp()
      }
    }
  }

  function * second() {
    try {
      if (position < index - 1) {
        firstCacheStart = position
        firstCache = new Array(index - position)
        for (let i = position; i < index; i++) {
          const item = iter.next()
          done = item.done
          if (!done) {
            firstCache[i] = item.value
          }
        }
      }

      while (!done) {
        const item = iter.next()
        done = item.done
        position++
        if (!done) {
          yield item.value
        }
      }
    } finally {
      cleanUp()
    }
  }

  return [first(), second()]
}
```

PR #182, opened 2019-02-03
Use jest projects and jest transformers

conartist6:
Changes the method we use to run jest on our sources when they have been transpiled to different language levels.

Jest is now runs much faster, and is capable of running the es2018 tests when run in an es2018 capable node version


sithmel:
I fixed a couple of minor things

PR #187, opened 2019-02-05
asyncSplitAt with optimization

KSXGitHub:


src/async-split-at.mjs
```diff
@@ -0,0 +1,69 @@
...
+    }
+  }
+
+  async function * second () {
+    try {
+      if (position < index - 1) {
+        firstCacheStart = position
+        firstCache = new Array(index - position)
+        for (let i = position; i < index; i++) {
+          const item = await iter.next()
```
conartist6: So you should put the promise of the next item into the cache, not the item itself.

conartist6: I could swear I had two comments which explained this, and only one of them showed up.

src/async-split-at.mjs
```diff
@@ -0,0 +1,69 @@
...
+  async function cleanUp () {
+    if (!done && typeof iter.return === 'function') {
+      await iter.return()
+    }
+  }
+
+  async function * first () {
+    try {
+      while (!done) {
+        if (firstCache) {
```
conartist6: It seems to me that you're suggesting that first needs to await before doing this check. That isn't the case. Second makes it synchronously when called.

src/async-split-at.mjs
```diff
@@ -0,0 +1,69 @@
...
+          yield firstCache[position - firstCacheStart]
```
conartist6: This line should be `yield await firstCache[...]`

src/async-split-at.mjs
conartist6: As long as you don't wait on the item promise before the item is in the cache, it will work

src/async-split-at.mjs
conartist6: Just think about it, it makes sense. The `await` needs to go in the loop because e still want `first` to be able to get each item as soon as it is ready.  And it's the most direct translation of what the logic was: whatever the iterator actually yielded, that's what belongs in the cache.

KSXGitHub:
Okay, now I'm really out of ideas with this one.

I not sure if `asyncSplitAt` needs optimization. The main use case of `splitAt` (sync version) is destructuring, async version exist mainly to fill the gap.

Anyway, I give up on this. @conartist6 @sithmel If you guys know how to fix the issue without complicating the code, you're free to try. Otherwise, I think async version that uses `asyncFork`/`asyncTee`

conartist6:
OK I fixed it. It was indeed a bit tricky. Let me write some better tests and I'll put up a PR

Issue #188, opened 2019-02-06
undefined should not be a valid iterable


sithmel:
In iter-tools we supported using as iterable null and undefined (both returning a sequence of 0 length).

In hindsight allowing undefined was a mistake. Two main reasons:

* it is an horrible way to swallow bugs
* undefined can be used for optional fields and interfere with the work for bringing consistency to all functions

conartist6:
As I said on the PR, I disagree strongly. A huge part of the benefit of using function utilities is that they work regardless of whether the value exists. This is not swallowing a bug, it is understanding that the code can be interpreted as written, and not creating a whole category of needless errors many of which are sure to be passed on to users of the software.

If some of our functions with trickier signatures still need custom currying logic in order to preserve behavior around undefined, that is a burden that we should be happy to bear in order to prevent a whole class of needless errors in code that uses iter-tools methods. It isn't the most elegant code that usually adds the most value, its the uglier code that saves anyone else from having to deal with the ugly shape of the problem.

conartist6:
I mean, just a core part of why iter-tools should be use widely is that you can do something like this without worrying if it's an error:
```js
map(val => val.toString(), myObj.undef)
```
Why should we demand that the user complicate this code? They'd have to write:
```js
map(val => val.toString(), myObj.undef || null)
```
but actually that's not really right at all because of 0 and false which should not be valid iterables, instead what really they should be writing *everywhere* in order to protect themselves is:
```js
map(val => val.toString(), myObj.undef === undefined ? null : myObj.undef)
// but likely that will be truncated to
map(val => val.toString(), myObj.undef == null ? null : myObj.undef)
```
And all of this gets even more needlessly complicated if you were doing something more like `getObj().undef`. Now what was a one line expression needs to be two lines to be correct, and most programmers are simply lazier than that.

conartist6:
I guarantee you though `|| null` is what will actually be written many places, which obscures real bugs and potentially creates incorrect runtime behavior

conartist6:
If we recommend the long form with `=== undefined` coders will shorten it to `== undefined` and then someone will see that and shorten it to `== null` and someone will see `myObj.undef == null ? null : myObj.undef` and they'll think that it must a no-op and remove it from the code and the code will still work for the moment because they won't happen to run a case where the value was undefined until even later on when something somewhere changes silently from returning undefined to null because the programmer didn't notice or care the difference and didn't think it could cause breakage far away in their code.

[...]

Issue #185, opened 2019-02-05
Consistency: call patterns and composability


sithmel:
I am picking up from #170 where the issue of consistency was raised. I'll try to classify the functions by calling arguments/return value. I am not listing all functions. But I think they are enough for a couple of observation and a proposal:

#### Fixed number of arguments, ended by an iterable/asyncIterable. Returns an iterable/asyncIterable:
* batch
* asyncBuffer
* cursor
* dropWhile
* flatMap
* flat
* interpose
* groupBy
* slice
* takeWhile
* tap
* asyncThrottle
* map
* filter

I like this calling pattern  It is simple enough and it enable function composition (using pipe or compose).

#### Fixed number of arguments, ended by an iterable/asyncIterable. Returns a value:
* every
* find
* some
* size
* consume ( returns undefined)

This is very similar to the one above. 

#### Fixed number of arguments, ended by an iterable/asyncIterable. Returns an iterable of iterables:
* partition
* fork
* splitBy (work in progress)
* splitAt (work in progress)

This is emerging recently. It enables to work with multiple slices (of clones) of the same iterable

#### Dynamic number of arguments
* asyncFilter
* asyncMap
* takeSorted 
* reduce (return a value)

For these I have a proposal. I am writing a curry that is based on the last argument (if it is an iterable it stops currying). This should sort these and the first list.

#### multiple iterables. return a single iterable
* concat
* compress
* zipAll
* zip
* merge/collate

For these I have a proposal. Given that we have a few functions returning iterable of iterables, it would ease function composition if these function would take an iterable of iterables as argument.

conartist6:
I can see the theoretical elegance of changing the functions which take multiple iterables to take an iterable of iterables, but it seems a bit like sacrificing the common use case for the benefit of the uncommon one.

KSXGitHub:
> I can see the theoretical elegance of changing the functions which take multiple iterables to take an iterable of iterables

Elegance in implementation, but not in usage.

> but it seems a bit like sacrificing the common use case for the benefit of the uncommon one.

I agree with this.

sithmel:
> I can see the theoretical ..

> .. like sacrificing the common use case for the benefit of the uncommon one ...

This is not theoretical or uncommon, this is how I write data pipelines!
I can also argue that this should be the preferred method.

Read this:
```js
const doubleAndSquare = (iterable) => {
  const [fork1, fork2] = fork(iterable)
  const doubles = map((x) => x * 2, fork1)
[...]

PR #189, opened 2019-02-06
Consistency fixes

sithmel:
This PR contains a few changes aimed to improve consistency in the library:

## new variadic curry function
Most of the functions are now using this wrapper. It helped to remove a lot of boilerplate:
* takes care of currying, and support variable arguments
* no need to wrap every iterable/asyncIterable in every function
* takes care of argument position. **Optional arguments go first**
* undefined is never a valid optional argument

## Breaking changes
~~undefined is not considered a "valid" iterable anymore~~
* takeSorted/asyncTakeSorted optional "comparator" argument is now the first argument


sithmel:
related tickets: #188 #185

KSXGitHub:
> undefined is not considered a "valid" iterable anymore

You meant "valid iterable element*"?

conartist6:
I don't think it's an acceptable trade off for our methods to fail on undefined. That's one of the absolute biggest benefits of iter-tools: it should protect you from the whole class of undefined/null errors.

CHANGELOG.md
```diff
@@ -14,6 +14,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Added
  - pipe
 
+### Changes
+ - takeSorted/asyncTakeSorted moved optional comparator argument to the beginning
```
conartist6: I recommend "optional comparator is now the first argument". Also I think you should also call out all the other arguments that can no longer be undefined here. My goal for this file is that if we change a method its name is in this file, such that if I search changelog for a method name I will know which releases have note-worthy changes to it.

sithmel: done

PR #190, opened 2019-02-09
splitAt and asyncSplitAt that work

conartist6:
I still need to write tests for return cases


src/__tests__/split-at.test.js
```diff
@@ -0,0 +1,105 @@
...
+  })
+
+  it('works when the source is exhuasted while the first half is being consumed', function () {
+    const [[a, b, c], [d, e, f]] = splitAt(3, slice(2, range()))
+    expect([[a, b, c], [d, e, f]]).toEqual([[0, 1, u], [u, u, u]])
+  })
+
+  it('works when the source is exhuasted while the second half is being consumed', function () {
+    const [[a, b, c], [d, e, f]] = splitAt(3, slice(4, range()))
+    expect([[a, b, c], [d, e, f]]).toEqual([[0, 1, 2], [3, u, u]])
```
KSXGitHub: Why don't you just use `undefined`? Using `u` alias may make the code looks pretty (which is a total coincidence), but it also makes the code harder to read.

src/__tests__/split-at.test.js
```diff
@@ -0,0 +1,105 @@
...
+    second.return()
+    expect([[a, b, c], [d, e, f]]).toEqual([[0, 1, 2], [3, 4, 5]])
+  })
+})
+
+describe('asyncSplitAt', function () {
+  it('works when the halves are consumed in order', async function () {
+    const [first, second] = asyncSplitAt(3, slice(6, range()))
+    expect(await asyncToArray(first)).toEqual([0, 1, 2])
+    expect(await asyncToArray(second)).toEqual([3, 4, 5])
```
KSXGitHub: You should combine 2 assertion lines above into one:

Instead of

```javascript
expect(a0).toEqual(a1)
expect(b0).toEqual(b1)
```

do

```javascript
expect({
  a: a0,
  b: b0
}).toEqual({
  a: a1,
  b: b1
})
```

[This is what I meant](https://github.com/sithmel/iter-tools/blob/b5fc2219d94cfef6bc6000f4eeb11ede4d32f069/src/__tests__/split-at.test.js#L9) (it is _array of arrays_ instead of _object of arrays_ but you get the idea)

**Why?** A test may fail multiple times. With separate assertions, we won't know if second assertion fails, but we should know with a single combined assertion.

**Alternate Solution:** Turn your "`it`" into 1 "`describe`" containing 2 "`it`"s

src/__tests__/split-at.test.js
conartist6: Haha what do you mean by "a total coincidence"? The point was to make it look nice. I think it is fairly self-evident from the string description, if nothing else, what `u` means. I can change it back to undefined too though, I do not really care.

src/__tests__/split-at.test.js
conartist6: Makes sense, I had been wondering why you did that.

src/__tests__/split-at.test.js
KSXGitHub: Well, I guess I should've said "a mere coincidence" instead (I'm not a native English speaker).

> The point was to make it look nice.

I already knew that, and "nice" is not a good reason.

KSXGitHub: To clarify further, the "coincidence" that I said is not "it happens to be pretty" but "the numbers  happen to have 1 character long the array happens to have 3 elements so you can use `u` to make it pretty"

src/__tests__/split-at.test.js
KSXGitHub: The more you know

src/__tests__/split-at.test.js
conartist6: I don't think you mean coincidence at all. If it was a coincidence that would mean that I did not intend to make it look nice.

KSXGitHub:
@conartist6 I have changed the title of the pull request, is this what you meant?

src/__tests__/split-at.test.js
conartist6: Ah OK I see what you mean. Yeah I know that it only looks nice for those reasons, but we do most of our tests with single digit numbers. But anyway, I'm not really meaning to make a debate out of this. I'll change it back.

conartist6:
Haha now we both changed that title. But yeah. Any thoughts on the code? I'm returning an iterable not an array now so the typedefs and tests need to be updated.

KSXGitHub:
> Any thoughts on the code?

It is pretty convoluted. I usually prefer simple code with mostly constant/immutable variables for it is easier to maintain and reason about. This is the opposite of my preference.

Any thought on this, @sithmel?

> I'm returning an iterable not an array now so the typedefs and tests need to be updated.

After the JavaScript part is settled, if the functions still return iterator of iterators, I'll take care of the types.

sithmel:
> It is pretty convoluted

It is indeed, but the best solution required a complex implementation. And I don't see how this can be made any simpler.

I like this. I'll merge as soon as the types are correct.

I am looking forward to splitBy :-D

sithmel:
Don't bother merging master into this branch. I am going to deal with that

sithmel:
@conartist6 @KSXGitHub I am happy to merge and finish this off (tests and @KSXGitHub comments) if you don't have time

conartist6:
@sithmel Yeah that's fine with me. My priorities are still with the new libraries I'm trying to get ready for release.

PR #191, opened 2019-02-11
some improvements on async-map

sithmel:
Added a few extra things:
* extended the same behaviour to async-flat-map
* fixed TS declaration for async-flat-map
* added return method to ParallelMapper
* some minor style change on ParallelMapper
* updated changelog
* updated readme


src/async-flat-map.mjs
```diff
@@ -1,11 +1,11 @@
 import map from './async-map'
 import { asyncIterableCurry } from './internal/async-iterable'
 
-async function * asyncFlatMap (func, iterable) {
+async function * asyncFlatMap (concurrency = 1, func, iterable) {
```
conartist6: This pattern is wrong -- a positional parameter default is used only if arguments.length is shorter than the position of the parameter.

src/async-map.mjs
```diff
@@ -13,74 +13,76 @@ class ParallelMapper {
...
-    }
+        if (previousUnresolvedPromise) {
+          // Ensure we don't emit parallelized items out of order
+          await previousUnresolvedPromise
+        }
+
+        if (item.done) {
+          return item
+        } else {
+          return { value: { value: item.value, mappedValue }, done: false }
```
conartist6: I forgot to touch this. There's no longer any need to emit and object including the value.

src/async-flat-map.mjs
sithmel: A default argument is used if you pass undefined. Example:
```js
const add = (a = 1, b) => a + b
add(3, 2) // 5
add(undefined, 2) // 3
```
asyncIterableCurry deals with passing undefined in the right place

src/async-map.mjs
sithmel: I am not sure I understand this

sithmel: Like:
```js
return { value:  mappedValue , done: false }
```
and 
```js
  for await (const value of runner) {
    yield value
  }
```
Am I correct?

sithmel: fixed

src/async-flat-map.mjs
conartist6: Just tested, you're right. The only distinction then is what happens if the user explicitly passes null, which seems less important when they can simply omit the argument.

PR #203, opened 2019-02-25
Upgrade curry for variadic functions. Fix tree shaking.

conartist6:


conartist6:
I still need to do a bit of testing. The location of the comment gets pretty funky after the transformation to CJS modules happens:
```
var _default = (
/*#__PURE__*/
0, _iterable.iterableCurry)(map);

exports.default = _default;
module.exports = exports["default"];
```

conartist6:
Oh and I need to trigger it for calls to the regular curry function to

conartist6:
Since the curry code is a bit heavier now, I pulled as much of it as can be fully parameterized out into a top level method so that it can be compiled and reused.

conartist6:
So I *think* that we don't really want or need a pure comment in the cjs files. I could change the configuration to only do that work for es module builds.

Rollup took over the work that I started on support `__#PURE__` comments (read: rewrote my diff) and it should ship soon.

Tree shaking is going to be a feature unique to users of either Rollup or Webpack, and both have default configurations which will preference our mjs over our js.

In particular we define `module` to point to an mjs file in our [package.json](https://github.com/sithmel/iter-tools/blob/ac293f486602e064bc9145f9e30668d9c29340fb/package.json#L6). [Rollup](https://github.com/rollup/rollup-plugin-node-resolve) will preference `module` over `main` by default, and so will [Webpack](https://webpack.js.org/configuration/resolve/#resolvemainfiles)

KSXGitHub:
> So I _think_ that we don't really want or need a pure comment in the cjs files. I could change the configuration to only do that work for es module builds.

That would be nice but you don't have to if it increases complexity.

conartist6:
@KSXGitHub: nope it's pretty easy. Done.

conartist6:
OK I'm content with it now, it should be ready to merge.

KSXGitHub:
> @KSXGitHub: nope it's pretty easy. Done.

Adding things will almost always be the easiest, the hard part is when you want to change or remove them.

conartist6:
@KSXGitHub do you have a specific worry?

KSXGitHub:
Nah, just clarify what I said in https://github.com/sithmel/iter-tools/pull/203#issuecomment-467982569 after reading [your response](https://github.com/sithmel/iter-tools/pull/203#issuecomment-467984462) (emphasis on "complexity" as opposed to "easy").

I also read your commits, it seems that [the last commit](https://github.com/sithmel/iter-tools/pull/203/commits/af9e648d60f837b493d176612c9fdf11df301272) adds no extra complexity, so I am fine with it so far.

src/internal/__tests__/curry.test.js
```diff
@@ -20,25 +20,28 @@ describe('curry', function () {
...
   it('throws with too many args', function () {
     const func = (a, b, c) => a + b + c
-    const f = variadicCurryWithValidation((x) => x === null, 'null', (x) => 0, func)
+    const f = variadicCurryWithValidation((x) => x === null, 'null', (x) => 0, func, false)
     expect(() => f(1)(2)(3)).toThrowError(new Error('func takes up to 2 arguments, followed by null. You already passed 3 arguments and the last argument was not null'))
   })
+  it('works on functions which take a variable number of iterables', function () {
+
+  })
 })
```
sithmel: this test is empty

babel-es2015.config.js
```diff
@@ -1,13 +1,14 @@
...
       plugins: [
-        'add-module-exports',
-        '@babel/plugin-transform-modules-commonjs'
+        "add-module-exports",
+        "@babel/plugin-transform-modules-commonjs"
       ]
     }
   }
-}
+};
```
sithmel: I think our linting rules use single quotes instead of double

conartist6:
@sithmel Fixed those things.

PR #205, opened 2019-03-02
call and apply

conartist6:


src/__tests__/apply.test.js
```diff
@@ -0,0 +1,22 @@
+/* eslint-env node, jest */
+const { apply } = require('..')
+
+describe('apply', function () {
+  let testFn = jest.fn()
```
KSXGitHub: Mocked functions should not be shared between multiple units. You should move this line to each `it` block instead.

sithmel:
At this point I don't see the necessity of having call/apply available.
The common case (my common case) can be solved with:

(_1, _2, ) => asyncZip(map(something, _1), filter(somethingElse, _2))

In case they are really needed they can be imported from ramda:
https://ramdajs.com/docs/#apply

I checked, apply can be curried:
https://ramdajs.com/repl/?v=0.26.1#?const%20nums%20%3D%20%5B1%2C%202%2C%203%2C%20-99%2C%2042%2C%206%2C%207%5D%3B%0AR.apply%28Math.max%29%28nums%29%3B%20%2F%2F%3D%3E%2042

src/__tests__/apply.test.js
```diff
@@ -0,0 +1,22 @@
...
+
+  beforeEach(function () {
+    jest.resetAllMocks()
+  })
+
+  it('passes the function the iterable of arguments provided to it', function () {
+    apply(testFn, [2, 3])
+    expect(testFn).toHaveBeenCalledTimes(1)
+    expect(testFn).toHaveBeenLastCalledWith(2, 3)
```
KSXGitHub: I always prefer _one assertion line per test unit_ (i.e. 1 `expect` per `it`). In this case:

```typescript
apply(testFn, [2, 3])
expect(testFn.mock.calls).toEqual([
  [2, 3]
])
```

src/index.d.ts
```diff
@@ -150,6 +150,9 @@ export declare function keys (obj: { [id: string]: any }): IterableIterator<stri
 export declare function values<T> (obj: { [id: string]: T }): IterableIterator<T>
 export declare function entries<T> (obj: { [id: string]: T }): IterableIterator<[string, T]>
 
+export declare function apply (fn: Function, args: Array<any>): any
+export declare function call (fn: Function, ...args: Array<any>): any
```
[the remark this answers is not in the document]
conartist6: I didn't, because typescript doesn't for `Function.prototype.call` or `Function.prototype.apply`. Tell me why it's not safe for them to use that type for the core function and I'll make this change.

src/apply.mjs
```diff
@@ -0,0 +1,9 @@
+import { curry } from './internal/curry'
+
+const emptyArray = []
+
+function apply (fn, args = emptyArray) {
```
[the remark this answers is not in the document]
conartist6: It's more concise, but it makes a different empty array every time, which we don't need.

src/__tests__/apply.test.js
conartist6: I don't care. I just simply do not begin to care. I'm sorry. You want to control the precise details of the code style, you write the feature.

conartist6:
@sithmel I was pretty sure that we just demonstrated powerful use cases for both `call` and `apply` as part of data pipelining. Also I don't think Ramda's `apply` accepts an iterable of arguments, and many people will want a dependency on either iter-tools *or* ramda, but likely not both. There's way too much overlap.

src/index.d.ts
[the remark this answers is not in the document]
KSXGitHub: `Function.prototype.call` and `Function.prototype.apply` are methods, their argument types depend on argument types of the callee. Thus, in order to define argument types of `call` and `apply`, `interface Function` has to have generic parameter as well, this would case a breaking change, so TypeScript decided not to change it.

**TL;DR** `Function.prototype.call` and `Function.prototype.apply` doesn't do it not because of safety, but because of early design limitation.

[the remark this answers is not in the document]
conartist6: OK, great explanation, thank you! I wasn't thinking about that. I will change the definitions. @sithmel doesn't this seem like another good value-add for these functions?

src/__tests__/apply.test.js
KSXGitHub: Dude, calm down. I do not know if you had a bad day, but "I don't care" is not really a resonable answer.

Whilst [my other suggestion](https://github.com/sithmel/iter-tools/pull/205#discussion_r261835609) might be mere code style, this one acts differently on failed tests.

src/apply.mjs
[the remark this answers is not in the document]
KSXGitHub: I don't think creating an empty array is any more expensive than mere integer allocation.

@sithmel Your opinion?

conartist6:
Integers aren't allocated unless I'm much mistaken.

sithmel:
@KSXGitHub your observations are correct, and thanks for adding links and explanations backing your claims.
The thing is that different styles can be both correct, depending on what kind of trade offs we are doing.
* Performance against readability
* terseness against easy to debug
* etc.

These discussions are very instructive but I am sure @conartist6 would like more useful feedbacks.

@conartist6 please keep the discussion constructive as usual

sithmel:
>  I was pretty sure that we just demonstrated powerful use cases for both call and apply as part of data pipelining. Also I don't think Ramda's apply accepts an iterable of arguments

Yes, that's true. I was wrong about suggesting Ramda.

There are 2 points that makes me lean to not include this in iter-tools.

## 1.
If I have to chose between:
```js
apply(zip)
```
and 
```js
([_1, _2]) => zip(_1, _2)
```
I would go for the arrow function. Apply used in this way would surprise the reader. But I admit, this is purely an opinion and I won't judge harshly the dev using **apply** instead.
With this in mind, given that you are already using an arrow function, I would use it only once and do:
```js
([_1, _2]) => asyncZip(map(something, _1), filter(somethingElse, _2))
```

## 2
Using an arrow function (instead of apply) is shorter and more clear
This is the biggest objection I have.

src/__tests__/apply.test.js
conartist6: My point was that it isn't a user facing distinction. It doesn't affect correctness or performance and it doesn't affect test correctness or performance. Since we often have rather intense discussion of how things should be designed or the structure of implementation, I try my best to avoid further nitpicking. We're just doing this because we enjoy doing it, and at the end of the day everyone will have slightly different ideas and patterns as long as we keep the differences away from the user that's OK.

Getting into the specifics of your suggestion, I don't find it to be particularly compelling because it is not internally consistent with other best practices. For example a competing best practice for the understandability of test debugging output is to use the most relevant possible matchers so that the debugging output is able to convey exactly what the expectation is. For example `toHaveBeenCalled` has better output than `toEqual`. This is a pattern that is broadly accepted across a variety of frameworks, all of which structure themselves to allow you to make multiple assertions. Basically if I wanted to wholeheartedly embrace the pattern you are suggesting everywhere, it would imply a different test framework than the one we are using.

Secondly I already took the practical step to reduce the risk of debugging confusion here, where is to put the more basic assumption first. If I make the call and the function isn't called at all, I don't see the point of proceeding further into the test like trying output the difference in arguments. That won't be why the test failed, it will have failed because the function wasn't called. So I would assert that my version already aligns with the intent of the framework we are using, and results in the most readable debugging output.

I'm sorry I didn't feel like writing this all out earlier.

conartist6:
@sithmel You can still do that in your code and suggest it in the pipelining docs, but `apply` still seems broadly useful. It's still far easier to read and write `apply(concat, iterables)` than
```js
iterables => concat(...iterables)(iterables)
```

sithmel:
I am ok fot the inclusion then. Let me review the code

src/apply.mjs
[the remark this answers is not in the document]
sithmel: they are both fine for me. Certainly not a blocker

CHANGELOG.md
```diff
@@ -15,6 +15,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
  - pipe
  - splitAt/asyncSplitAt
  - multiPartition/asyncMultiPartition
+ - collate, asyncCollate: Formerly named merge. No longer takes a helper. Default with no params is now interleaving
```
sithmel: I don't think this PR includes any change on merge or collate

src/__tests__/pipeline.test.js
```diff
@@ -9,9 +9,21 @@ function isEven (x) {
...
     const iter = pipeline([1, 2, 3, 4], map(bump), filter(isEven))
     expect(Array.from(iter)).toEqual([4, 6])
   })
+
+  it('can create a complex data processing pipeline', function () {
+    expect(Array.from(pipeline(
+      [3, 3],
+      fork,
+      slice(2),
+      merge(apply).partial([add, mul])
```
sithmel: partial?
* I thought we discarded the idea of partial method
* I don't see it implemented, how can this pass the test ?

src/__tests__/apply.test.js
KSXGitHub: I guess you have your point. But I still insist on [this](https://github.com/sithmel/iter-tools/pull/205#discussion_r261835886)

src/__tests__/apply.test.js
KSXGitHub: @conartist6 You still have this to be resolved. Mocked functions contain side-effects, which should **never** be shared by multiple independant tests.

sithmel: @KSXGitHub this should take care of it:
```js
  beforeEach(function () {
    jest.resetAllMocks()
  })
```
I am still quite new to jest, so I don't know if this is really an issue.

KSXGitHub: Resetting all mocks is not only unnecessary, it would even cause logic error.

Resetting just `testFn` alone is okay, but I still think side-effects should not be shared.

sithmel: > it would even cause logic error.

how so?

KSXGitHub: > > it would even cause logic error.
> 
> how so?

It won't right now, but it will when we add more mocks to the test.

KSXGitHub: Even if it never does, I still prefer a more straightforward code. Imagine reading through all the `beforeEach`/`beforeAll`/`afterEach`/`afterAll` and try to reason with it! I personally never use unnecessary `beforeEach` unless I have to make side-effects (such as modifying filesystem, etc.)

sithmel: I see your point about making the code more straightforward. From the context I don't see how this can really make any difference. But it is such a small change and I am ok doing it.

CHANGELOG.md
conartist6: Oops, that's a bit of the bigger PR that snuck into this one

src/index.d.ts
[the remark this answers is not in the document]
conartist6: @KSXGitHub Do you happen to know what the lowest version of Typescript is that we support with our typedefs is? Typescript 2.8 has the `ReturnType` type, but I'm not sure if using that would reduce our support.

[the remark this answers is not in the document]
KSXGitHub: 3.0 since https://github.com/sithmel/iter-tools/pull/55. And I don't think this should concern us to be honest, very few (if none) uses old TypeScript versions.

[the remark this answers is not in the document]
KSXGitHub: And BTW, why do you need `ReturnType`?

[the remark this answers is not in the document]
conartist6: Well the typedefs that you had me write seem to be part of what is now the `ReturnType` type, so I thought maybe we should just use that. I don't know why they do `infer R : R` though, and we don't have that bit.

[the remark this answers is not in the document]
KSXGitHub: > Well the typedefs that you had me write seem to be part of what is now the `ReturnType` type

Oh, you were talking about that `Return` generic parameter? That is just a name, you can change it to anything (such as `R`, `T`) and it still works. I just chose `Return` because it is a return type.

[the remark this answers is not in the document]
KSXGitHub: BTW, since don't know what built-it `ReturnType` is for (as you said "I don't know why they do `infer R : R` though"), I'm gonna explain it to you:

It is to infer return type of a function, for example:

```typescript
declare function myFunc (): 'foo' | 'bar'
type R = ReturnType<typeof myFunc> // R should be 'foo' | 'bar'
```

And `infer` is used to infer, for example:

```typescript
type ArrayElement<Arr> = Arr extends Array<infer X> ? X : any
type R = ArrayElement<[0, 1, 2, 3]> // R should be 0 | 1 | 2 | 3
```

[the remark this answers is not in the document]
conartist6: Typescript [says](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-8.html):
"For example, the following extracts the return type of a function type:"
```ts
type ReturnType<T> = T extends (...args: any[]) => infer R ? R : any;
```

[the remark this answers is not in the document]
conartist6: I guess that's because the ReturnType type is used when you need to infer from another type. When you're extracting a generic type from a parameter it's obvious that you're "inferring".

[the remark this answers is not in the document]
KSXGitHub: That's what I said

CHANGELOG.md
```diff
@@ -15,6 +15,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
  - pipe
  - splitAt/asyncSplitAt
  - multiPartition/asyncMultiPartition
+ - mergeAll, asyncMergeAll
```
sithmel: This is a future feature.
But I can fix it for you @conartist6 
Is this ready to be merged ?

conartist6: Thanks that would be great. I'm working on the `interleaveGenerator` now. I do believe this is ready to go.

PR #206, opened 2019-03-03
interleave

conartist6:


src/__spec__/interleave-generator.spec.ts
```diff
@@ -0,0 +1,130 @@
...
+>(
+  iter.interleaveGenerator(
+    function * (
+      canTakeAny: Function,
+      buffer: iter.InterleaveBuffer<string>
+    ) {
+      yield buffer.take()
+    }
+  )(['foo'])
+)
```
KSXGitHub: You should add more to this test:

```typescript
assert<
  IterableIterator<string>
>(
  iter.interleaveGenerator(
    function * (
      canTakeAny, // no needs for type here, so that you can test if TypeScript can figure it out (type inference)
      buffer // no type here, same reason as above
    ) {
      // test type of buffer
      assert<iter.InterleaveBuffer<string>>(buffer)

      // expand 'yield buffer.take()'
      const res = buffer.take()
      assert<string>(res) // to test typeof buffer.take()
      yield res // to test if buffer.take() fits in return type

      // add test for buffer.canTake()
      assert<boolean>(buffer.canTake())

      // add test for canTakeAny()
      assert<string>(canTakeAny())
    }
  )(['foo'])
)
```

src/interleave-generator.mjs
```diff
@@ -0,0 +1,42 @@
...
+    if (next.done) {
+      return undefined
+    }
+
+    this._next = this._iterator.next()
+    return next.value
+  }
+
+  canTake () {
+    return this._next.done ? null : this
```
KSXGitHub: Wait, I thought `buffer.canTake()` returns `boolean`? Your TS declaration also says it returns `boolean`!

src/interleave-generator.mjs
```diff
@@ -0,0 +1,42 @@
...
+  }
+}
+
+export default function interleaveGenerator (generatorFn) {
+  return (...iterables) => {
+    return (function * () {
+      const iterators = map(iterable => ensureIterable(iterable)[Symbol.iterator](), iterables)
+      const buffers = Array.from(map(iterator => new InterleaveBuffer(iterator), iterators))
+      const canTakeAny = () => buffers.find(buffer => buffer.canTake())
```
KSXGitHub: I am not sure if `canTakeAny` is an appropriate name since it implies returning `boolean` ("can \<something\>?" is a yes/no question)

src/interleave-generator.mjs
```diff
@@ -0,0 +1,42 @@
...
+    const next = this._next
+    if (next.done) {
+      return undefined
+    }
+
+    this._next = this._iterator.next()
+    return next.value
+  }
+
+  canTake () {
```
KSXGitHub: **[DISCUSSION]** Should `canTake` be a property getter (or a property) instead of a function? Because tagged union would prove to be useful in TypeScript:

```typescript
// declaration

interface InterleaveBufferThatCanTake<T> {
  canTake: true
  take (): T
}

interface InterleaveBufferThatCannotTake {
  canTake: false
}

type InterleaveBuffer<T> = InterleaveBufferThatCanTake<T> | InterleaveBufferThatCannotTake
```

```typescript
// usage

if (buffer.canTake) {
  yield buffer.take() // no errors
} else {
  yield buffer.take() // TSC: Property 'take' does not exist on type 'InterleaveBufferThatCannotTake'
}
```

KSXGitHub: On a second thought, this would probably not necessary, but TS declaration need to be tweaked:

```typescript
// declaration

interface InterleaveBuffer<T> {
  canTake (): this is InterleaveBufferThatCanTake<T>
}

interface InterleaveBufferThatCanTake<T> extends InterleaveBuffer<T> {
  take (): T
}
```

```typescript
// usage

if (buffer.canTake()) {
  yield buffer.take() // no errors
} else {
  yield buffer.take() // TSC: Property 'take' does not exist on type 'InterleaveBuffer<T>'
}
```

src/interleave-generator.mjs
conartist6: Ah. Yeah I need to update the types and docs. There's a couple reasons it makes sense to do this instead of boolean. In the async variant it saves some really gross code. It also permits doing clever things like `buffer.canTake()?.take()`. You can only do that though if the type system tells you what the type really is.

KSXGitHub: I think `boolean` is more straightforward. As for the async variant, why don't you just return a promise of (sync) `InterleaveBuffer` (`Promise<InterleaveBuffer<T>>`) instead of making an `AsyncInterleaveBuffer<T>` where each method returns a promise?

src/interleave-generator.mjs
conartist6: I struggled with that same thought for a while too. I decided to name it based on its primary intended usage. I think what I aught to do is when it is being used in a different way my example code should bind it to a different name, like `findTakableBuffer`. I have the one example like that at the bottom of the docs for the async version.

src/interleave-generator.mjs
conartist6: I don't want to do that because I also decided not to make it an error to `take()` when `canTake()` is falsy. Basically if you don't check `canTake` only `canTakeAny` your interleave will end up filling missing values with `undefined`, which we do other places like `zipLongest` anyway. I believe that some people will want to do that. The tricky thing from the Typescript perspective though is that we don't capture the circumstance in which undefined may be introduced to the result type, so from that perspective I can see the attractiveness of using the types to force the user into a usage that the compiler can understand. I do think it's OK to have code that is more permissive in untyped usage than typed usage. Curious to hear what you think about this.

src/interleave-generator.mjs
conartist6: That won't work because the buffers are passed in as arguments. We'd pass a different promise for every item, which means we'd have to pass a function `getBufferPromise()` which I think is not cleaner than what the code does now.

src/interleave-generator.mjs
KSXGitHub: Since the type of the final iterable does not contain `undefined`, I would like TypeScript to prevent me from fill the iterable with hidden `undefined` values.

conartist6: The thing is I don't want the error that the type system gives to be that the method `take` does not exist. That is actively misleading. Is there a way to express the idea that calling the function is a type error without hiding the whole function?

src/async-interleave-generator.mjs
```diff
@@ -0,0 +1,43 @@
...
+  async canTake () {
+    return (await this._nextPromise).done ? null : this
+  }
+}
+
+export default function interleaveGenerator (generatorFn) {
+  return (...iterables) => {
+    return (async function * () {
+      const iterators = map(iterable => ensureAsyncIterable(iterable)[Symbol.asyncIterator](), iterables)
+      const buffers = Array.from(map(iterator => new AsyncInterleaveBuffer(iterator), iterators))
```
KSXGitHub: Why don't you just use `Array.prototype.map`? That would make the code faster, more concise, and lighter

```javascript
const buffers = iterables
  .map(iterable => ensureAsyncIterable(iterable)[Symbol.asyncIterator]())
  .map(iterator => new AsyncInterleaveBuffer(iterator))
```

src/interleave-generator.mjs
KSXGitHub: But that would make it even more misleading IMO, so it is best to make it disappear.

KSXGitHub: You can also make it returns `unknown` (3.2 feature) when `canTake()` return `true`.

KSXGitHub: Okay, this is final: Return `T | undefined` when `canTake()` is `false`.

```typescript
// declaration

interface InterleaveBuffer<T> {
  canTake (): this is InterleaveBufferThatCanTake<T>
  take (): T | undefined // or null (I'm too lazy to double-check your code)
}

interface InterleaveBufferThatCanTake<T> extends InterleaveBuffer<T> {
  take (): T
}
```

This way, TSC will only screams error when `strictNullCheck` is set to `true`, just like the way TypeScript was designed.

conartist6: I just came to the same conclusion myself. https://www.typescriptlang.org/play/index.html#src=interface%20EmptyInterleaveBuffer%3CT%3E%20%7B%0D%0A%20%20%20%20canTake()%3A%20this%20is%20InterleaveBuffer%3CT%3E%0D%0A%20%20%20%20take%20()%3A%20undefined%20%7C%20T%0D%0A%7D%0D%0A%0D%0Ainterface%20InterleaveBuffer%3CT%3E%20extends%20EmptyInterleaveBuffer%3CT%3E%20%7B%0D%0A%20%20take%20()%3A%20T%0D%0A%7D%0D%0A%0D%0Adeclare%20var%20buffer%3A%20EmptyInterleaveBuffer%3Cstring%3E%3B%0D%0A%0D%0Aconst%20taken%20%3D%20buffer.take()%3B%0D%0A%0D%0Aif%20(buffer.canTake())%20%7B%0D%0A%20%20%20%20const%20takenSafely%20%3D%20buffer.take()%0D%0A%7D

KSXGitHub: I don't think "Empty" is necessary, a true `EmptyInterleaveBuffer` would have `take()` returns only `undefined`.

conartist6: Now the problem is that I still don't really want the return value of `canTake` to be a boolean for the reasons discussed previously. How bad do you think it would be if the type just misrepresents what is returned a bit?

conartist6: Also this approach just won't work for async.

KSXGitHub: Then we should redesign `InterleaveBuffer`. Instead of having just one `class InterleaveBuffer`, we should have two:
  * One that provides `canTake` method that returns an object that provides `take`
  * One that provides `take`

And since both classes have just one method, we should return functions directly instead.

src/__spec__/interleave-generator.spec.ts
conartist6:
```
Argument of type 'IterableIterator<{}>' is not assignable to parameter of type 'IterableIterator<string>'.
  Type '{}' is not assignable to type 'string'.
```

conartist6: It is inferring `InterleaveBuffer<{}>` for some reason

KSXGitHub: I'll look into this tomorrow (which, to my timezone, is probably until 6-10 hours later).

KSXGitHub: In the meantime, you can explicitly add type annotation to where it causes error.

src/interleave-generator.mjs
conartist6: If you leave Typescript out of the picture that is a worse API design than what we have. I don't want to create API designs that type systems will never be able to understand, but I also don't feel obligated to twist my APIs up because the current feature set of a type system doesn't give users the tools to verify that certain kinds of things are safe.

src/interleave-generator.mjs
KSXGitHub: Okay, so you want `buffer.canTake()?.take()` which is just a slightly shorter version of `buffer.canTake() && buffer.take()`. I don't this is compelling enough of a reason to abandon TS checker.

Your async argument is a good point though, but I wouldn't do a `(await buffer.canTake())?.take()` for it looks pretty cluttered.

Beside, user doesn't want `undefined` in their resulting iterable for most of the times, so they would still use `if` anyway.

src/interleave-generator.mjs
conartist6: I'm going to mark this resolved as I have updated this location with the type safety solution we discussed. We can still continue the discussion about how it should be handled for the async version, which I have left as it was.

src/internal/race-to.mjs
```diff
@@ -0,0 +1,23 @@
+export default function raceTo (predicate, notFoundValue, promises) {
```
sithmel: This one might deserve a test.

src/internal/race-to.mjs
```diff
@@ -0,0 +1,23 @@
...
+            resolved = true
+          }
+        })
+        .finally(() => {
+          settledPromiseCount++
+          if (!resolved && settledPromiseCount === promiseCount) {
+            resolve(notFoundValue)
+          }
+        })
+      promiseCount++
```
sithmel: You have the number of promises already. Why counting them ?

src/internal/race-to.mjs
```diff
@@ -0,0 +1,23 @@
+export default function raceTo (predicate, notFoundValue, promises) {
+  return new Promise(resolve => {
+    let promiseCount = 0
+    let settledPromiseCount = 0
+    let resolved = false
+    for (const promise of promises) {
+      promise
```
sithmel: Instead of looping through them, you can wrap all promises in a promise race.

sithmel:
I have added a couple of comments.
interleaveGenerator is a nice helper to combine together multiple iterables. And I think that helps to implement a family of new functions in a more readable way.

I have a few questions:
* I am not sure about the name. It does interleave. But not strictly alternating between iterables. Also unsure about generator. Maybe interleaveIterableFactory ?
* are we sure we want to expose this outside? it looks a bit weird given that is the only function with that kind of interface
* do you still want to reimplement zip/zipAll with this?

src/internal/race-to.mjs
conartist6: I have an iterable of promises, so I am counting how many promises I got in the iterable, and then I am also counting how many have resolved in one way or another.

conartist6:
To your questions:

I still plan to use it internally. It will used to implement `collate` (interleave by comparator), `interleave` (round robin), and `asyncInterleaveReady` (if you still want that).

I'm a bit flexible on the name. Perhaps the prefix `custom` would denote the level of its flexibility better? I wanted `generator` as part of the name to cue the user in to the fact that they need to plug a generator function into it. I should also do a better job in the README of explaining right up front what you would be trying to accomplish when you need this function, and needing to plug a generator in.

Finally I want the name to look good as a decorator since I anticipate that the community will eventually demand decorator support for plain functions:
```js
@interleaveGenerator
function * interleaveAABB(canTakeAny, a, b) {
  while (canTakeAny) {
    if (a.canTake()) yield a.take()
    if (a.canTake()) yield a.take()
    if (b.canTake()) yield b.take()
    if (b.canTake()) yield b.take()
  }
}
```

As to whether it should be exposed at all, well, do you think that looking a bit weird carries a risk?

sithmel:
> I still plan to use it internally. It will used to implement collate (interleave by comparator), interleave (round robin), and asyncInterleaveReady (if you still want that).

I still want those :+1: 

> As to whether it should be exposed at all, well, do you think that looking a bit weird carries a risk?

the fact that is exposed should not be a problem. How you present it, it might lead to confusion.

We might present it as "interleaveHelper(generator, ...iterables)" and that would sits a bit more naturally between the other functions. You can still use it as a decorator because of the curry. But you can also use it like any other function. What do you think ?

src/internal/race-to.mjs
sithmel: oh yes, because it is an iterable, not an array.
Maybe it would be more practical convert the iterable to array though. (see comment on promise.race)

src/internal/race-to.mjs
sithmel: > I don't want to resolve if an input promise rejects

maybe you should let this blown up if it rejects

And you lost me with the rest of explanation :-)

Maybe you can just go on implementing the test, and then I can fiddle with it. So I will either understand how it works or propose an alternative

src/async-interleave-generator.mjs
```diff
@@ -0,0 +1,44 @@
...
+  }
+}
+
+export default function interleaveGenerator (generatorFn) {
+  return (...iterables) => {
+    return (async function * () {
+      const buffers = iterables.map(iterable =>
+         new AsyncInterleaveBuffer(ensureAsyncIterable(iterable)[Symbol.asyncIterator]())
+      )
+      const canTakeAny = () => raceTo(Boolean, false, map(async buffer => (await buffer.canTake()) ? buffer : null, buffers))
```
KSXGitHub: I see this:

```javascript
map(async buffer => (await buffer.canTake()) ? buffer : null, buffers)
```

Why don't you just use `buffers.map`?

KSXGitHub: I also have another question:

So `raceTo` returns a **promise** that resolves an object (`Promise<AsyncInterleaveBuffer>`) that has 2 methods (`canTake` and `take`) that return promise again?

This would require user to do at least 3 `await` in callback function, one for the `AsyncInterleaveBuffer`, one for `canTake` and one for `take`.

Why don't you pass a wrapper instead? i.e.

```javascript
// your old class
class AsyncInterleaveBuffer { /* ... */ }

// the wrapper
const createCanTakeAny = (...args) => () => {
  const bufferPromise = raceTo(...args)
  return {
    async canTake () {
      return (await bufferPromise).canTake()
    },
    async take () {
      return (await bufferPromise).take()
    }
  }
}
```

This way, user wouldn't have to `await` to get `AsyncInterleaveGenerator`

conartist6: Why would I use buffers.map? It creates an array I don't need to allocate and it would create that array for every item in every iterable.

conartist6: To your second question, that's right. I don't think the overhead of writing await a few times is worth paying any significant price to fix, and I think what you are proposing has cost. It has perf cost, and it is likely to either be different from the sync form of the API or unnecessarily complicate it.

src/internal/race-to.mjs
conartist6: I just meant that you can't say that `canTakeAny` is true if all you do is race to the first iterable that is ready to tell you that `canTake: false`. I probably should make sure it blows up in canTake ever actually rejects though.

src/internal/race-to.mjs
conartist6: Again this is in a pretty tight loop. If I don't have to make an array, I won't. Inside tight loops are where iterators are most helpful.

src/async-interleave-generator.mjs
KSXGitHub: > It has perf cost

It's always going to be, either from the code of `iter-tools`, or from the user who is going to use `await` to turn the promise into something useful.

> and it is likely to either be different from the sync form of the API

It already different from the sync form (e.g. `buffer.take()` return a promise that resolves a value instead of the value directly).

conartist6:
I'm ready to merge this. @sithmel any final thoughts?

sithmel:
I thought we decided to make it an internal function, and to use to implement all others "merge" functions

conartist6:
Oh, maybe we did. I couldn't remember where we left this so I was just looking through the comments on this thread.

I'd still like to make it public. My intent was to offer a replacement for the public merge API that I still plan to deprecate. It'd be nice to be able to tell anyone who was using that that they should use this instead.

conartist6:
I know it's a bit different from other things in the library, but I think anyone who sees an example is gonna grok the concept pretty quick.

sithmel:
I don't think is the right level of abstraction for iter-tools. This was the main regret I had for "merge", it required to fiddle with merging detail.

I would rather deal with the common cases.
- merge by position (sync/async)
- merge by value (sync/async)
- merge by time (async)

conartist6:
Remind me what you mean by "merge by position"?

conartist6:
Just a round robin collate right...

sithmel:
possibly in 5 distinct function ;-)

I think the best thing for now is to make interleave generator "internal" and merge.

conartist6:
Could we call it "experimental" instead of internal? We could make an section of the docs for experimental methods and call this one `experimentalInterleaveGenerator`.

We could link each method to an issue e.g. `promote experimentalInterleaveGenerator to interleaveGenerator` and request that people who want to see the method promoted thumbs up the initial description, while anyone who feels the API is not well designed could give a thumbs down. We could also encourage people to share their use cases in the issue comments.

conartist6:
Anything with no thumbs or use cases we'd delete or make internal as appropriate after some length of time.

sithmel:
I don't see why we should leave the design of the library to some passer-by. Surely they won't be involved enough to make a thoughtful decision.

If this is really important to you, maybe you can move in the utility section. I really wouldn't want to expand the library with a family of decorator.

Do you have any other decorators in mind by any chance ?

conartist6:
I absolutely do think there will be other decorators. Two of them exist already in some form: throttle, and buffer.. 

Eventually I see these being used as:
```js
@throttle(10)
function * myThrottledGenerator(throttledSource) { ... }
// and
@buffer(2)
function * myBufferedGenerator(bufferedSource) { ... }
```

I imagine we'll come up with more too. If you think these all fall outside the scope of the library, then I can create somewhere else to publish them.

conartist6:
The cool thing is that because the `asyncThrottle` and `asyncBuffer` functions are curried, they already work like this (or they will as soon as js supports decorators on non-class-method functions).

conartist6:
Actually now that I look at it it seems fairly unlikely to be that throttle and buffer will be used that way. A consumer needs to be responsible for buffering and throttling, and in that example those concerns are on the producer of values, which doesn't make much sense.

sithmel:
Shall we go with the utility section for now ?

conartist6:
I realize I'm being thick-skulled here. This doesn't need to be `interleaveGenerator` at all, it can just be `interleave`. As I've pointed out, our currying syntax already facilitates usage of our regular functions as decorators, if that ever becomes appropriate.

Once you take that away, `interleave` looks very much like every other method in the library, except that it expects you to pass a generator function in place of a normal callback.

conartist6:
@sithmel Would such a change allay your fears about introducing this concept of decorators and widening the definition of what belongs in the library?

sithmel:
> Would such a change allay your fears about introducing this concept of decorators (no longer needed) and widening the definition of what belongs in the library?

yes :-)

Last thing.
We are now back with something similar to what "merge" is.

I understand you found the name "merge" confusing. Are you sure of "interleave" ?

But I won't insist on this, if you can't come up with anything better, I am happy with interleave (I couldn't figure out a better name, anyway).

conartist6:
I like merge. Merge can be mistaken conceptually with spreading as in `Object.merge`, and since the previous method was called merge it's an added bonus that this one, which works quite differently, does not share the same name.

conartist6:
Now I just to add the uncurried version of the typedefs

conartist6:
Added uncurried typedefs.

PR #207, opened 2019-03-05
Fix interleaveGenerator related types

KSXGitHub:
Here is some temporary fixes, there're so much to be done, I will improve this in the future.

**Changes:**
  * Removed async type tests (I'll reintroduce them later when I tough the async types)
  * Exported the interfaces (TS user may need them)
  * Rewrote interleaveGenerator function signatures
  * Rewrote sync type tests

**Related:** https://github.com/sithmel/iter-tools/pull/206#discussion_r262145559


src/__spec__/interleave-generator.spec.ts
```diff
@@ -4,127 +4,83 @@ import * as iter from '../index'
...
-      yield b2.take()
-      yield b3.take()
-    }
-  )(['foo'], [2], [(_: any) => _])
-)
+      for (const buffer of buffers) {
+        assert<string | undefined>(buffer.take())
+        assert<boolean>(buffer.canTake())
 
-assert<
```
conartist6: You can't just remove these. Sad though it is, when we have to make five overloads of the type to account for situations with different parameters, we need five tests, one for each variant. I think it's important to keep these tests because it can be particularly hard to spot subtle differences in those highly duplicated type overloads.

src/__spec__/interleave-generator.spec.ts
```diff
@@ -4,127 +4,83 @@ import * as iter from '../index'
...
+        const res = b2.take()
+        assert<number>(res)
+        yield res
+      }
     }
-  )(['foo'], [2], [(_: any) => _])
+  )(['foo'], [2])
 )
-
-assert<
```
conartist6: Why would we not include any type test? We should test as much of type as we know works, even if some more complex type functionality does not yet.

src/index.d.ts
```diff
@@ -473,22 +473,38 @@ export declare function zipAll<T> (...iterables: Array<Iterable<T>>): IterableIt
...
-export declare function interleaveGenerator<T1, T2, T3> (gen: (canTakeAny: () => InterleaveBuffer<T1 | T2 | T3> | null, b1: InterleaveBuffer<T1>, b2: InterleaveBuffer<T2>, b3: InterleaveBuffer<T3>) => any): (i1: Iterable<T1>, i2: Iterable<T2>, i3: Iterable<T3>) => IterableIterator<T1 | T2 | T3>
-export declare function interleaveGenerator<T1, T2, T3, T4> (gen: (canTakeAny: () => InterleaveBuffer<T1 | T2 | T3 | T4> | null, b1: InterleaveBuffer<T1>, b2: InterleaveBuffer<T2>, b3: InterleaveBuffer<T3>, b4: InterleaveBuffer<T4>) => any): (i1: Iterable<T1>, i2: Iterable<T2>, i3: Iterable<T3>, i4: Iterable<T4>) => IterableIterator<T1 | T2 | T3 | T4>
+export type InterleaveCallback<Args extends InterleaveBuffer<any>[]> = (
+  canTakeAny: () => IterableElement<Args> | null,
+  ...buffers: Args
+) => Iterable<
+  Args extends Iterable<InterleaveBuffer<infer X>> ? X : never
+>
+
+export declare function interleaveGenerator<T1, T2> (
```
conartist6: Huh OK that cleaned up pretty nicely.

src/__spec__/interleave-generator.spec.ts
KSXGitHub: This is just glorified version of `assert<string, number>`, I don't see this as necessary.

src/__spec__/interleave-generator.spec.ts
KSXGitHub: Your criticism is valid. However, as I said in the original comment, this is only temporary, I plan to replace the many overloads with a single one that covers all use cases.

src/__spec__/interleave-generator.spec.ts
conartist6: Huh? It's tests the async defs.

src/index.d.ts
KSXGitHub: OK. Whenever you want to comment on a block, please choose the last line of the block so that I can see the whole block that you commented.

src/__spec__/interleave-generator.spec.ts
KSXGitHub: > It tests the async defs.

Ah, then as I already said in the original comment, I have taken the async tests out until I touch it again.

src/__spec__/interleave-generator.spec.ts
conartist6: How is that possible?

KSXGitHub: Everything seems impossible until we actually try.

conartist6: Hahaha OK then. I'm just saying that all the type code I've seen uses multiple overloads to avoid a design limitation of typescript: that its spread type is unable to infer a tuple type, forcing instead an array like `Array<string|number>`. While that's not exactly wrong as a type it allows you to put the parameters in the wrong positions. I don't see how you'll get around this, but I'll be pretty impressed if you manage it.

KSXGitHub: I can manage this, but there's a catch: User must always specify generic parameters, like this:

```
const generator = interleaveGenerator<[string, number, boolean]> (function * (
  canTakeAny,
  someString,
  someNumber,
  someBoolean
) { /* ... */ })
```

I'll do it if you're OK with this.

conartist6: Unless I'm much mistaken the types I wrote allow you to do that without any changes at all, don't they? And the user doesn't always have to specify the generic parameters: if they prefer they can eithe import InterleaveBuffer and specify the types of the parameters to the function in terms of `InterleaveBuffer<T>`.

If I just make my `T_` types default to `any` the really ugly error messages go away too when I forget to specify types using either method, and I'm just left with standard warnings about having a function with untyped parameters.

Let me draw up those changes and improve the type tests and see what you think.

KSXGitHub: In my current implementation, user doesn't always have to specify generic parameters, but they must specify type annotations of callback function.

conartist6:
I have finally rebased my interleave PR on top of my latest changes, and I have incorporated the improvements suggested in this PR.

PR #208, opened 2019-03-05
Newgroupby

sithmel:
## WIP: Don't merge, you can start review!

This is a proposal of a different interface for the Exchange object.

It also contains a new implementation of groupBy that buffers the original iterable. This allows to use groupBy with destructuring.

Together with groupBy refactoring I am experimenting a splitBy function that should be able to slice "horizontally" a sequence by any criteria, efficiently.


sithmel:
I reimplemented splitAt using splitBy.
splitBy can be used split an iterable in slices. It keeps the space efficiency of splitAt but is more versatile.

Currently missing:
- more tests
- TS definitions
- documentation

An example in the cookbook on how to slice an array in multiple slices (I have an interesting example in mind).

Apart the missing things, what is in the PR should be fine, so I am happy if anyone can have a look and give feedback.

src/group-by.mjs
```diff
@@ -1,39 +1,99 @@
...
-      } else {
-        currentItem = iterator.next()
+  }
+
+  // generate group returning [key, subgroup]
+  // it always picks up new items from the Iterator (fetch)
+  function * generateGroups () {
+    // using an empty object as initial key:
+    // it is surely different from any possible key
+    let currentKey = {}
```
conartist6: I think you could just call this `key`. I also don't really mind if you leave it this way though. I have just begun to think that `current` is usually meaningless. If you just say `item` and `key` it will be obvious that the current item is whatever is in the variable. And you can still have `nextItem` or `previousItem` of course.

Also I think self documenting code techniques would work here too. You could make this a constant like `const UNIQUE_INITIAL_KEY = {}`.

src/internal/queues.mjs
```diff
@@ -44,19 +29,34 @@ export class Queue {
...
-    if (!this.head) throw new Error('You cannot spawn a new consumer after setting calling noMoreConsumers')
-    return new Consumer(this.head)
+export class Exchange extends Queue {
+  shift () {
+    throw new Error('Unsupported')
   }
 
-  noMoreConsumers () {
-    // this enables to garbage collect all the consumed items
+  getConsumer () {
```
conartist6: I think with the change to consumers cloning each other, the exchange base class no longer makes much sense. What about just having it be a function used like this: `const { consumer, cacheValue /* fn(value) */ } = createExchange();`?

src/__tests__/split-by.test.js
```diff
@@ -0,0 +1,20 @@
+/* eslint-env node, jest */
+const { splitBy, asyncSplitBy, asyncToArray } = require('..')
+
+describe('splitBy', function () {
+  it('splitBy', function () {
+    const [a, b, c] = splitBy('AAABBCCCCD')
```
conartist6: Would be super helpful as a reviewer to have at very least a test that demonstrates the basic use case. It documents the function for me, and when I see the green flag from Travis I know that the code I'm reading actually does what it claims to.

src/split-by.mjs
```diff
@@ -0,0 +1,93 @@
+import { iterableCurry } from './internal/iterable'
+import { Exchange } from './internal/queues'
+
+function splitBy (getKey = (k) => k, iterable) {
```
conartist6: Couldn't you replace this whole thing with `map(([,group]) => group, groupBy)`

But also what happened to the splitBy we discussed which had a predicate callback that identified separator items which would not be present in the final output?

conartist6:
The new group by implementation looks good though. I very much like that it is internally consistent with the perf characteristics of our other methods like splitAt and fork.

src/fork.mjs
```diff
@@ -26,13 +26,13 @@ export default function fork (iterable) {
     }
   }
 
-  function * generateFork (a) {
+  function * generateFork (cons) {
```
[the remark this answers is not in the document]
sithmel: It surely can't be worst that "a" :-)

src/__tests__/split-by.test.js
sithmel: yes, sorry. I won't solicit feed backs anymore without tests

src/internal/queues.mjs
conartist6: Actually no that won't work because you still need `noMoreConsumers()` to be part of the API, which you have gotten rid of here. We put that in for forking infinite iterables. You can't unallocate the exchange because you still need to be able to push items into it. Exchange also can't keep its head, because that forces the GC to keep everything in the linked list, even items that have been consumed by every consumer.

conartist6: Ignore my previous comment, it was wholly wrong.

conartist6: Ah damn, no, it wasn't wholly wrong. This breaks fork. We need to add a new test to fork. The case which breaks is:
```js
forks = fork(range())
forks.next().value.next() // 1
forks.next().value.next() // 2, but should be 1.
```

conartist6: Wait, I ran that as a test on your branch and it passed. My head hurts.
It passes because you are keeping the original consumer to clone from, which mean that the bug is far more subtle. Keeping that consumer means that you are preventing any cache data from ever being GC'd.

sithmel:
I don't think you are seeing a bug.
How you treat  the "main consumer" of the queue is key here.

In fork the main consumer is scoped in the generator function emitting forked iterables: generateForks

That means that when this generator is exhausted, the main consumer should be unreachable and therefore garbage collected.

In groupBy and splitBy is different. The main consumer is always in the main function scope. But in this case it should never hold more than one item in the queue: in the fetch function, when I insert a new item in the queue I always move the main consumer forward. So it will point to the next item inserted.

src/split-by.mjs
sithmel: groupBy and splitBy are similar, but also different.

In splitBy every single iterator emitted is responsible to consume the original iterator.

In groupBy, there is a "main loop" that looks ahead to search for different keys.

This also changes the output of asyncGroupBy. In fact this returns an asyncIterable of asyncIterables while asyncSplitBy returns an iterable of asyncIterables

src/internal/queues.mjs
sithmel: This can't be done because, as I explain [below](https://github.com/sithmel/iter-tools/pull/208#issuecomment-471233351) I need to create the consumer in a specific place, and not necessarily together in the same scope as the main object.

src/group-by.mjs
sithmel: I can't use key, because I have many other keys around, and it would be confusing,

I used currentKey, because this is the key of the main loop that is currently in use.

I like the idea of using a constant UNIQUE_INITIAL_KEY :+1:

conartist6:
OK so I see now what you're saying about the main consumer being garbage collected. I now agree that there is no bug. I still don't love it from the perspective of long term maintenance because it takes logic which nobody is going to understand implicitly and makes it very implicit.

Anyway, thinking there was a bug (and also just because I needed to do this in order to really understand what I was looking at) I tweaked it a bit and came up with my own version. Take a look and see if there's anything that you like.

https://gist.github.com/conartist6/e3e7e19fde884b920fca7359163640ec

src/group-by.mjs
```diff
@@ -1,39 +1,97 @@
 import { iterableCurry } from './internal/iterable'
+import { Exchange } from './internal/queues'
 
-function * groupBy (key = (k) => k, iterable) {
+const UNIQUE_INITIAL_KEY = {}
+
+function groupBy (getKey = (k) => k, iterable) {
   const iterator = iterable[Symbol.iterator]()
```
conartist6: This is a bug actually. Never noticed it before. We can't open the source iterator until we're inside the try block with the finally that cleans it up.

src/split-by.mjs
conartist6: I'm really still not understanding. How can it synchronously know how many groups it is going to create out of an asynchronous iterable? Or, looking at the test that exist, it requires the caller to know how many groups they expect? That seems like an API flaw to me more than a feature. And conceptually it is still grouping.

I could get the result easily enough (for sync things) then like this:

const [[, As], [, Bs] = groupBy("AABB")

sithmel: > How can it synchronously know how many groups it is going to create out of an asynchronous iterable?

It doesn't. Just like fork.
If you are interested in working with all groups you will have to consume one at the time until it returns an empty group.

conartist6: Why would this be a better API than the one we had discussed though?

conartist6: I'm working on an implementation of splitBy that matches String.prototype.split functionality. I think it is well worthwhile not to force the user to handle the termination condition themselves, which would make it virtually impossible to chain the method.

src/group-by.mjs
sithmel: I don't think this is a problem.
If you have never called "yield" on the iterator, you won't need to clean it up.

src/split-by.mjs
sithmel: > Why would this be a better API than the one we had discussed though?

I designed splitBy to be a low level implementation for any case where you slice an iterable. You can build any other case on top of it (as I did for splitAt).

sithmel: > make it virtually impossible to chain the method

it seems to be a bit an overstatement to me. I am sure there are plenty of cases where this would do the trick.

src/group-by.mjs
conartist6: That's not true. It is true of our iterators, but not of iterators or iterables in general.

sithmel: can you explain better?
```js
const iterator = iterable[Symbol.iterator]()
```
should return an iterator or an exception and they are both fine.

You need to call
```js
iterator.return()
```
only after the first call to
```js
iterator.next()
```

src/split-by.mjs
conartist6: My biggest complaint is that it doesn't seem to have anything to do with splitting. It's a low level grouping function. To split something, you must be able to choose the point at which to split it. That is either a size 1 match (item => splitOnItem) or a size 0 match (a, b) => splitBetweenAandB. To group you define a characteristic for the group, and items sharing the characteristic are part of the group, which is what this function is about.

That would mean that we still need a splitting method, which I have written [here](https://gist.github.com/conartist6/e3e7e19fde884b920fca7359163640ec#file-split-by-mjs).

src/group-by.mjs
conartist6: Remember iterators aren't always generator functions. Someone could give you an iterable over a file:
```js
class FileIterator {
  constructor(name) {
    this.handle = fs.openSync(name)
  }
  next() { /* yield line from file */
  return() {
    fs.closeSync(this.handle)
  }
}

class File {
  [Symbol.iterator]() {
    return new FileIterator(this.name)
  }
}

Array.from(new File('./myFile'))
```

conartist6: So what you say about needing to call `return` only after the first call to `next` really isn't true. It is usually OK in our code because we usually open external iterators on our first iteration, not before.

src/split-by.mjs
conartist6: If you want, I'll make the async version and turn that gist into a PR on this PR

src/group-by.mjs
sithmel: That would be an incorrect use of the iterator interface. And it won't work correctly in a for..of loop either.

src/split-by.mjs
sithmel: > My biggest complaint is that it doesn't seem to have anything to do with splitting

We can still come up with a different name.

> If you want, I'll make the async version and turn that gist into a PR on this PR

No please wait for it.
As I said I'd like to use this as a low level abstraction for these family of functions. I don't want to maintain 3/4 functions that are almost the same.

conartist6: Yeah it would be great if these functions could share some more code. I can work on trying to get my implementations to share some, but to do that I'd first need to know that you want a split function with that API.

conartist6: I just don't want to work backwards from internal code sharing to the external API

sithmel: The way you implemented splitBy is very similar to groupBy. The main loop consumes the original iterator (this is why you are able to know when to stop producing iterators).

This way, however, you are unable to reimplement asyncSplitAt, because you can't do this:
```js
const [iter1, iter2, iter3] = asyncSplitBy(iterable) // asyncIterable can't be spread
```

That is exactly the reason why I think both groupBy and splitBy (the way I implemented, replace with any name you prefer ) covers 2 different and valuable use cases.

conartist6: Couldn't splitAt be implemented with plain old `groupBy` though?
```js
function splitAt(index, iterable) {
  const [[, first]. [, second]] = groupBy((_, i) => i < index ? 0 : 1, iterable)
  return [first, second]
}
```

conartist6: Ah but you're saying specifically that that doesn't work for async, I see.

sithmel: yep

conartist6: But that's still pretty easy to get around.
```js
function asyncSplitAt(index, iterable) {
  const groups = groupBy((_, i) => i < index ? 0 : 1, iterable)
  let firstGroup = groups.next();
  let secondGroup = groups.next();
  async function * first() {
    yield * await firstGroup
  }
  async function * second() {
    yield * await secondGroup()
  }
  return [first(), second()]
}
```

sithmel: "await" in a sync function ?

conartist6: Yeah my bad see revised version.

sithmel: Let me think. If we can suppress splitBy, in favour of groupBy, I am more than happy to do so

conartist6: I guess in theory we could implement splitBy with groupBy as well I suppose?
```js
splitBy(predicate, iterable) {
  const grouped = groupBy((item, i) => !!predicate(item, i), iterable)
  return filter((_, i) => isEven(i), grouped)
}
```

src/group-by.mjs
conartist6: If you believe that would be an incorrect use of the iterator interface, could you point me to where it is specified that that is not allowed?

sithmel: https://tc39.github.io/ecma262/#sec-iterator-interface

> The returned object must conform to the IteratorResult interface. Invoking this method notifies the Iterator object that the caller does not intend to make any more next

"any more" next implies there was at least one. Does that convince you ?

It is perhaps never stated explicitly, but this is how they implemented generator functions. This is the heuristic used by:
* for .. of
* array spread
* etc.

If someone implements things differently it is already broken using js primitives.

sithmel:
@conartist6 In the latest commit I implemented your idea of getConsumer returning the the tail of the queue instead of the head (but without creating an extra class).
Now there is no risk of creating implicit memory leaks. it made me also save some lines of code !

src/group-by.mjs
conartist6: I like your line of thought, but in fact you are disproved by the array spread operator, which will call return even in the unusual case that it does not attempt to spread any items.

https://jsfiddle.net/dyqp54vr/

src/group-by.mjs
```diff
@@ -1,39 +1,96 @@
...
+        // or
+        // currentItem.key === currentKey that means that currentItem
+        // a fetch called in a subgroup returned an item of a different key.
+        if (!currentItem || currentItem.key === currentKey) {
+          fetch()
+        }
+
+        if (done) return
+
+        if (currentItem.key !== currentKey) {
```
conartist6: I would like a better comment here (or nearby). Actually I think a lot of the current comments don't tell much more than the code itself does, but here especially is where I got tripped up.

This line is the core of how/why your implementation works, and I only just fully understood it now. It causes the main loop to have a double purpose (triple?) purpose. It yields groups, the fetch handles one-item lookaheads (to tell if the next item be part of this group), and the combination of those two also serve the purpose of fetching the entire previous group, but only if it has not already been fetched as a result of the code on the other side of the `yield` consuming the result of `generateGroups()`. I understand now that our implementations are effectively equivalent in their primary technique and behavior.

src/internal/queues.mjs
```diff
@@ -44,19 +29,37 @@ export class Queue {
...
-  spawnConsumer () {
-    if (!this.head) throw new Error('You cannot spawn a new consumer after setting calling noMoreConsumers')
-    return new Consumer(this.head)
+  push (data) {
+    const newItem = new QueueItem(data)
+    this.tail.previous = newItem
+    this.tail = newItem
   }
 
-  noMoreConsumers () {
```
conartist6: I still rather liked `noMoreConsumers`. Sure it was a bit cryptic and long as a name, but I think that's exactly what you want for a method that begs to reader to go look carefully at what is happening. It also created a spot where detailed documentation could be coupled to the smallest chunk of code that wholly encapsulated the desired behavior.

src/split-by.mjs
conartist6: That won't work though if you have two separators next to each other, in which case you should get an empty group. Anyway I still think this method is premature, and it's the last thing stopping me from saying that you should go ahead and merge this.

src/group-by.mjs
sithmel: Alternative idea. How are you going to call "return" in the instance? if the constructor fails you don't get the instance at all.

src/group-by.mjs
sithmel: You are in the ideal position to suggest the best possible comment here :-)

src/internal/queues.mjs
sithmel: The current design makes *noMoreConsumers* redundant and I am pretty happy with it.
It's not clear what you ask here

src/group-by.mjs
conartist6: I'm just saying the code should be:
```js
let iterator
...
function * generateGroups() {
  try {
    iterator = iterable[Symbol.iterator]()
  } finally {}
}
```
I don't know what you're talking about in terms of the constructor failing.

sithmel: with the empty finally block ?

conartist6: Uh, no? Keep everything the way it is now. Just move the initialization inside finally. Or don't, and I'll open a PR once this is merged.

sithmel:
> That won't work though if you have two separators next to each other, in which case you should get an empty group. Anyway I still think this method is premature, and it's the last thing stopping me from saying that you should go ahead and merge this.

Yes, that's the same thing stopping me. I'll have a thought about it.

sithmel:
I have built a proof of concept of *split* built using splitBy:
```js
function * cons (item, iterable) {
  yield item
  yield * iterable
}

function carcdr (iterable) {
  const iterator = iterable[Symbol.iterator]()
  const {done, value} = iterator.next()
  if (done) return []
  return [value, iterator]
}

function * split (separator, iterable) {
  const isSeparator = item => item === separator
  const subseqs = splitBy(isSeparator, iterable)

  for (const subseq of subseqs) {
    const [first, rest] = carcdr(subseq)
    if (first === undefined && rest === undefined) return
    if (!isSeparator(first)) {
      yield cons(first, rest)
    } else {
      const [second, restRest] = carcdr(rest)
      if (!isSeparator(second)) {
        continue
      } else {
        yield map(() => '', cons(second, restRest))
      }
    }
  }
}
```
The example includes a couple of utilities inspired by scheme (cons, carcdr). I like using them, but I am unsure others will find clear/useful.

sithmel:
Using similar pattern to implement groupBy with splitBy:
```js
  function * groupBy (getKey, iterable) {
    for (const subseq of splitBy(getKey, iterable)) {
      const [first, rest] = carcdr(subseq)
      if (first === undefined && rest === undefined) return
      const key = getKey(first)
      yield [key, cons(first, rest)]
    }
  }
```
Note: I can get groupBy out of splitBy, but I can't do other way around. This is because splitBy is lazier, while groupBy eagerly consumes the iterable searching for the next key.

sithmel:
## proposal
* removing the old groupBy and rename splitBy as groupBy.
* adding an entry in the cookbook on how to implement the old groupBy

Once this is done I don't think is necessary implement other functions of this family (except splitAt that is already implemented).

conartist6:
Here's the biggest problem: in your splitAt implementation every group gets the head of the list (`exchange.getConsumer`) instead of getting a consumer which starts reading where the last group stopped. This gives the algorithm n squared performance, which is not acceptable. You did this intentionally in order to dodge async-ness of getting groups, so I don't see how the concept can survive.

sithmel:
This new implementation of Exchange returns a new consumer starting with the tail (previously was the head). And this: ```starts reading where the last group stopped```

> This gives the algorithm n squared performance, which is not acceptable. You did this intentionally in order to dodge async-ness of getting groups, so I don't see how the concept can survive.

I don't know why this has anything to do with the running time, and I don't understand how this has anything to do with async-ness

conartist6:
Oh right I forgot that you did that. If it was the old version spawning each consumer at the head, it would impact running time as I suggest.

If you're going to propose a groupBy based on splitBy, I would like the opportunity to see tests and poke at it.

Looking at your proposed function I see some problems. It does a check `first === undefined`. The iterable may contain undefined, so I'd think this part of the condition should be dropped, which seems like it would be fine.

Also your cons implementation seems bad. It's a linked list whose nodes are generator functions linked with closures. It's novel, certainly, but not good code.

sithmel:
> If you're going to propose a groupBy based on splitBy, I would like the opportunity to see tests and poke at it.

Now, you have it

sithmel:
I would like to move this forward, so I can start taking care of other things (there are plenty).
I am aware that this is still missing tests, definitions and documentation. But before that I would like to be sure I am treading in the right direction.

I feel happy/confident about the splitBy implementation and the way it is used to efficiently reimplement splitAt and groupBy. Also I don't want to implement some opinionated version of "split" that works like String.prototype.split. Apart using that with strings, I am not sure is that useful and it can be implemented starting with splitBy anyway (an example in the cookbook can be a good idea).

The only thing I am not 100% sure about is to call the new function splitBy. But I don't dislike it too much, so I can still consider it an option.

conartist6:
Apologies I've been crazy busy. Could you agree to just make split by
internal for the moment? It doesn't mean we can't release it. Let's just
merge this so we can move forward with other discussions and PRs and once
we have everything else together maybe it will be more clear how to name
and market the function correctly known as splitBy.

conartist6:
*currently known

sithmel:
No need to apology, same here.
I moved splitBy in internal as proposed.
I think this is ready to merge

PR #210, opened 2019-03-20
Proposal: Allow multiple iterations

conartist6:
The idea is to make the Iterable aspect of our IterableIterators return separate iterator instances.

Reasoning: we require iterables as input, and many iterables can be iterated over more than once (e.g. Arrays, Maps, and Sets). If you have an iterable that can be consumed multiple times and you use any of our functions on it, you are left with an iterable that can only be consumed once.

~~In theory we could continue to support both multiple consumption and IterableIterators, but since `Symbol.iterator` is almost always going to be called to get the iterator which will actually be used (and must always return a fresh iterator), this leaves us creating a "static" iterator to be the iterator part of IterableIterator, and it will usually never be consumed.~~

~~I think there are some specific cases where it still makes sense to return IterableIterators, particularly for inifinite iterables where it may be desirable to get values one at a time with `next`.~~

EDIT: Originally I proposed changing IterableIterators to Iterables. I am no longer proposing that, so the above is no longer relevant.

All method tests now pass. Curry helper tests need some rethinking. Skipped async merge test because I don't know why it failed and we're planning to get rid of that method prior to the next release anyway.


conartist6:
Of course we could get tricky and only instantiate a real static iterator the first time next was called. Actually I don't see why we wouldn't do this. Then the typedefs don't need to change at all, and only 4 suites fail. Updating the PR.

conartist6:
Why is it important that it be possible to iterate more than once? After all, why not just make a cache with an array at that point?

1. Infinite iterables. When the iterable in question is infinite it cannot be cached in an array. It could be cached with fork, but it many cases that will be a significantly more expensive operation than just recomputing the data.

2. Debugging. It's highly useful to be able to evaluate what the value of some iterable transformation is, and at present doing this will destroy the program. While this is a nuisance for human debuggers, it is also problematic for writing warnings, which may want to check an iterable for abnormal values (when in dev or CI environments) and yet not alter further execution of the program (or it's perf characteristics in prod).

3. Use as transformation. If your source iterable is mutable, (like Map, Set, or Array) operations like `map` may have different results on different executions. This is actually highly desirable, and allows the iterable to serve as a "transformation" of the source which unlike an allocated data structure never needs additional work to be kept in sync.

4. Long lived sources. It may not be helpful to cache the result of an iterable transformation to a static source if it is likely to be a long time between iterations over the result. The cost/benefit ratio is maximized for a simple transformation (e.g. `map(val => val.id)`) over a large data set when the result is needed infrequently.

src/internal/curry.mjs
```diff
@@ -17,12 +17,49 @@ function unshiftUndefineds (args, by) {
...
+  [Symbol.iterator] () {
+    return this._iterate()
+  }
+}
+
+class AsyncIterable extends BaseIterable {
+  [Symbol.asyncIterator] () {
+    return this._iterate()
+  }
+}
```
KSXGitHub: So instead of returning simple iterators, you create instances of these classes? My problem with this is that this is quite bloated, not to mention that babel often adds extra overhead for class inheritance, this would make JavaScript bundle size larger.

But I also think that you can make it more lightweight by converting class inheritance to object composition.

KSXGitHub: Actually, I think you don't need "composition" part, only "object". What `BaseIterable.constructor` does is merely storing arguments, if you replace these classes with a factory function, you won't be needing those 7 lines of constructor function, you also won't have to use `this`.

conartist6: You are right that the library is most sensitive to bundle size that babel's class transpilation would add here, and that the overhead of classes is not necessary. I have changed this to prototypal inheritance. Does this resolve your concern?

src/internal/curry.mjs
```diff
@@ -17,40 +17,47 @@ function unshiftUndefineds (args, by) {
...
-    this._iterableArgs = iterableArgs
-    this._staticIterator = null
-  }
+function BaseIterable (fn, variadic, args, iterableArgs) {
+  this._fn = fn
+  this._args = args
+  this._variadic = variadic
+  this._iterableArgs = iterableArgs
+  this._staticIterator = null
+}
```
KSXGitHub: I think you can make it even shorter. (this is merely converting ES6 classes to ES5 constructor functions, and you still use `this`)

Anyway, if you still don't know what I meant, I'll write it tomorrow.

conartist6: If you intend to replace an instance with some pattern involving closures I think you will be sacrificing runtime perf at the expense of a very small amount of bundle size. For the most core part of the libraries it is important that we create functions which can get hot and be compiled to native code. This means that the dynamic things we work with should be objects, not functions, and should have a stable shape.

KSXGitHub:
I've taken a look at the code. It seems that you achieve this by creating an abstraction that would call the actual function over and over again, correct?

So I offer another approach: Instead of having these abstraction by default, we let user chooses. We would create a wrapper function called `eternal` (or `reusable`, `everlasting`, etc.) that would return either a reusable iterable or a function that return a reusable iterable.

```javascript
// this still makes an iterator
const a = range(4)

// this makes a reusable iterable
const b = eternal(range, 4)

// this makes a function
const f = eternal(range)
const c = f(4)

// a is not reusable
const a1 = Array.from(a) // [0, 1, 2, 3]
const a2 = Array.from(a) // []

// b is reusable, and so is c
const b1 = Array.from(b) // [0, 1, 2, 3]
const b2 = Array.from(b) // [0, 1, 2, 3]
```

conartist6:
Um, I can't think of any reason we would not embrace the iterable spec itself. In the iterable spec `[Symbol.iterator]` is a factory function for iterators. Why would we force the user to use some separate API to create a in iterable whose `[Symbol.iterator]` is a factory function for iterators?

src/internal/curry.mjs
KSXGitHub: It sounds like a premature optimization to me, and it requires you to predict (or assume) the internal working of all JavaScript engines, which are unreliable and always changing. And you don't even have a proper benchmark.

The only thing that I am certain of is bundle size and readability, and I find the closure approach satisfies both.

That is not to say that performance is unnecessary. It is. But without benchmark, we only take a shot in the dark.

KSXGitHub:
> Why would we force the user to use some separate API to create a in iterable whose `[Symbol.iterator]` is a factory function for iterators?

So that we won't force user to always spent bundle size on these curry abstractions. Also, implementing `eternal` should be more lightweight, like this:

```javascript
const eternal = (fn, ...args) => ({ [Symbol.iterator]: fn(...args) })
```

src/internal/curry.mjs
conartist6: May I suggest that instead of throwing your arms up and declaring that it is impossible to understand how javascript engines work and that they having nothing in common, you do some studying on the topic? I recommend you start [here](https://mathiasbynens.be/notes/shapes-ics) and continue with [this](https://mathiasbynens.be/notes/prototypes)

conartist6:
I'm not particularly worried about the contribution to the bundle size. It is a relatively small addition for what it gives us, the sources will almost always be gzipped before going over the wire/air (making it smaller), and most people are going to be using more than one function from the library, making the percentage increase in bundle size negligible.

src/internal/curry.mjs
KSXGitHub: Your provided links do not invalidate my arguments which is:
  1. Avoid premature optimization
  2. Engines are always changing

Beside, the links you provided all point to a single V8 engine. There're also SpiderMonkey, JavaScriptCore, etc. That is one of the reasons why benchmark is needed when performance is concern.

And what not to say reducing bundle size does not gives you performance benefit? Download time, parsing time.

conartist6: The links note the similarities between many engines, not just v8. They *all* have optimizing compilers. Out of curiosity, do you know C or C++? Have you dealt with structs and pointer arithmetic, for example? There is much that is not always changing about the way that computers work. It is by writing code that mirrors the low level way a processor actually does work that you avoid the uncertainty of how various JS engines treat the more dynamic patterns that javascript allows.

conartist6: But what is certain is that if you create a function many times, even if it seems like one function in your mind, it is not to the compiler. The compiler cannot optimize repeated calls to the same function unless the repeated calls are actually to the same function. This is basic stuff, not compiler internal rocket science. There *may* be some compiler internal rocket science that can reverse engineer that multiple methods really should be the same, but that would be the kind of thing that we would be always changing and we would definitely not want to rely on.

conartist6: I'm going to reverse your argument on you though. You're talking about making sacrifices to correctness to gain 1KB of raw bundle size, and less when compression is in use! That's premature optimization if I've ever heard of it.

conartist6:
Furthermore your proposed eternal function is unnecessarily toxic to static analysis. If indeed this is a pattern that should be embraced in order to embrace it using your proposed function would require the user to forgo ever actually calling the functions they intend to use, instead using an `apply`-like syntax.

src/internal/curry.mjs
conartist6: > "Premature optimization" is a phrase used to describe a situation where a programmer lets performance considerations affect the design of a piece of code. This can result in a design that is not as clean as it could have been or code that is incorrect, because the code is complicated by the optimization and the programmer is distracted by optimizing.

(From https://en.wikipedia.org/wiki/Program_optimization#When_to_optimize)

Can we agree on this? Correctness first, optimization second? Not to say that we should accept solutions which are the wrong complexity class or code which is twice as slow as it should be, but correctness should always come first, and should be *followed* by optimization if necessary.

sithmel:
@conartist6 I'd like to have a look and do some experiment, but tests are failing.

@KSXGitHub I don't understand your concern about using classes. I checked on the babel interactive repl:
```js
class Test {
  constructor(a, b) {
    this.a = a
    this.b = b;
  }
  method1() {
    return this.a + this.b
  }
  method2() {
    return this.a + this.b
  }
  method3() {
    return this.a + this.b
  }
}
```
chars 192

gets translated in:
```js
"use strict";

var Test =
/*#__PURE__*/
function () {
  function Test(a, b) {
    this.a = a;
    this.b = b;
  }

  var _proto = Test.prototype;

  _proto.method1 = function method1() {
    return this.a + this.b;
  };

  _proto.method2 = function method2() {
    return this.a + this.b;
  };

  _proto.method3 = function method3() {
    return this.a + this.b;
  };

  return Test;
}();
```
363 chars

Here's the equivalent using prototypes:
```js
"use strict";

function Test(a, b) {
  this.a = a;
  this.b = b;
}

function Test.prototype.method1 () {
  return this.a + this.b;
}

function Test.prototype.method2 () {
  return this.a + this.b;
}

function Test.prototype.method3 () {
  return this.a + this.b;
}
```
247 characters

and using a factory function:
```js
"use strict";

function Test(a, b) {
  return {
    function method1 () {
      return a + b
    }
    function method2 () {
      return a + b
    }
    function method3 () {
      return a + b
    }
  }
}
```
192 characters

Take into consideration that:
* factory functions can return different objects and this can't be optimised by any js engine. @KSXGitHub  You are right that browser engines are many, but you have to consider that v8 is the biggest player, and optimisations techniques are more or less the same across different engines
* factory function returns a bigger objects, because the functions are all closures with a different surrounding environment
* the class syntax overhead can be counted in bytes. After minification, we are probably looking at less than 100 bytes
* the prototype function implemented is much harder to read than the original class

I propose to go back to the original class based implementation

KSXGitHub:
> I don't understand your concern about using classes

My concern was not classes themselves, but class inheritance, babel adds a lot of overhead in order to achieve this. This can be solved by either:
  * Use prototype inheritance (Current implementation)
  * Use class composition
  * Use factory functions

> factory functions can return different objects and this can't be optimised by any js engine

Your speculation cannot convince me, only benchmarks can.

sithmel:
> a lot of overhead

I tried to quantify how much overhead. It didn't seems to be "a lot"

> Your speculation cannot convince me, only benchmarks can.

https://coderwall.com/p/jj6fwa/the-prototype-is-your-friend-if-you-care-about-perf

KSXGitHub:
@sithmel OK. You convinced me.

conartist6:
Failing es5 tests are caused by https://github.com/babel/babel/issues/9755

conartist6:
@sithmel just run the es2015 tests if you want to experiment. I'll update the tests for the curry helper itself now.

sithmel:
@conartist6 I don't know what I am missing. I am getting weird errors in the test:
```
 FAIL  src/__tests__/partition.test.js
  ● Test suite failed to run

    Cannot find module '..' from 'partition.test.js'

       8 |   asyncToArray,
       9 |   asyncIterable
    > 10 | } = require('..')
         |     ^
      11 | 

```

conartist6:
@sithmel I don't think any of my changes should be responsible for causing that error. Have you done an `npm install` and `npx jest --clearCache`?

sithmel:
@conartist6 I tried what you suggested. I even tried to clone it again in a new directory. Update to the latest node 10.
If I run ```npm test``` I get a lot of errors like this:
```
 asyncTap › return tapped iterable (curried version)

    TypeError: Object is not async iterable

      1 | import { ensureAsyncIterable } from './internal/async-iterable'
      2 | 
    > 3 | export default async function asyncToArray (iterable) {
        |                ^
      4 |   const out = []
      5 |   for await (const item of ensureAsyncIterable(iterable)) {
      6 |     out.push(item)

      at _asyncIterator (node_modules/@babel/runtime-corejs2/helpers/asyncIterator.js:22:9)
      at _callee$ (src/async-to-array.mjs:3:16)
      at tryCatch (node_modules/regenerator-runtime/runtime.js:62:40)
      at Generator.invoke [as _invoke] (node_modules/regenerator-runtime/runtime.js:288:22)
      at Generator.prototype.(anonymous function) [as next] (node_modules/regenerator-runtime/runtime.js:114:21)
      at asyncGeneratorStep (node_modules/@babel/runtime-corejs2/helpers/asyncToGenerator.js:5:24)
      at _next (node_modules/@babel/runtime-corejs2/helpers/asyncToGenerator.js:27:9)
      at node_modules/@babel/runtime-corejs2/helpers/asyncToGenerator.js:34:7
      at new F (node_modules/core-js/library/modules/_export.js:36:28)
      at node_modules/@babel/runtime-corejs2/helpers/asyncToGenerator.js:23:12
      at asyncToArray (src/async-to-array.mjs:19:24)
      at Object.asyncToArray (src/__tests__/tap.test.js:39:18)
      at tryCatch (node_modules/regenerator-runtime/runtime.js:62:40)
      at Generator.invoke [as _invoke] (node_modules/regenerator-runtime/runtime.js:288:22)
      at Generator.prototype.(anonymous function) [as next] (node_modules/regenerator-runtime/runtime.js:114:21)
      at asyncGeneratorStep (node_modules/@babel/runtime-corejs2/helpers/asyncToGenerator.js:5:24)
      at _next (node_modules/@babel/runtime-corejs2/helpers/asyncToGenerator.js:27:9)
      at node_modules/@babel/runtime-corejs2/helpers/asyncToGenerator.js:34:7
      at new F (node_modules/core-js/library/modules/_export.js:36:28)
      at Object.<anonymous> (node_modules/@babel/runtime-corejs2/helpers/asyncToGenerator.js:23:12)

 FAIL  src/__tests__/take-sorted.test.js
  ● asyncTakeSorted › return smallest iterable

    TypeError: Object is not async iterable

      1 | import { ensureAsyncIterable } from './internal/async-iterable'
      2 | 
    > 3 | export default async function asyncToArray (iterable) {
        |                ^
      4 |   const out = []
      5 |   for await (const item of ensureAsyncIterable(iterable)) {
      6 |     out.push(item)

      at _asyncIterator (node_modules/@babel/runtime-corejs2/helpers/asyncIterator.js:22:9)
      at _callee$ (src/async-to-array.mjs:3:16)
      at tryCatch (node_modules/regenerator-runtime/runtime.js:62:40)
      at Generator.invoke [as _invoke] (node_modules/regenerator-runtime/runtime.js:288:22)
      at Generator.prototype.(anonymous function) [as next] (node_modules/regenerator-runtime/runtime.js:114:21)
      at asyncGeneratorStep (node_modules/@babel/runtime-corejs2/helpers/asyncToGenerator.js:5:24)
      at _next (node_modules/@babel/runtime-corejs2/helpers/asyncToGenerator.js:27:9)
      at node_modules/@babel/runtime-corejs2/helpers/asyncToGenerator.js:34:7
      at new F (node_modules/core-js/library/modules/_export.js:36:28)
      at node_modules/@babel/runtime-corejs2/helpers/asyncToGenerator.js:23:12
      at _asyncToArray (src/async-to-array.mjs:112:24)
      at asyncToArray (src/async-to-array.mjs:19:24)
      at Object.asyncToArray (src/__tests__/take-sorted.test.js:24:18)
      at tryCatch (node_modules/regenerator-runtime/runtime.js:62:40)
      at Generator.invoke [as _invoke] (node_modules/regenerator-runtime/runtime.js:288:22)
      at Generator.prototype.(anonymous function) [as next] (node_modules/regenerator-runtime/runtime.js:114:21)
      at asyncGeneratorStep (node_modules/@babel/runtime-corejs2/helpers/asyncToGenerator.js:5:24)
      at _next (node_modules/@babel/runtime-corejs2/helpers/asyncToGenerator.js:27:9)
      at node_modules/@babel/runtime-corejs2/helpers/asyncToGenerator.js:34:7
      at new F (node_modules/core-js/library/modules/_export.js:36:28)
      at Object.<anonymous> (node_modules/@babel/runtime-corejs2/helpers/asyncToGenerator.js:23:12)

  ● asyncTakeSorted › return smallest iterable, using comparator

    TypeError: Object is not async iterable

      1 | import { ensureAsyncIterable } from './internal/async-iterable'
      2 | 
    > 3 | export default async function asyncToArray (iterable) {
        |                ^
      4 |   const out = []
      5 |   for await (const item of ensureAsyncIterable(iterable)) {
      6 |     out.push(item)

      at _asyncIterator (node_modules/@babel/runtime-corejs2/helpers/asyncIterator.js:22:9)
      at _callee$ (src/async-to-array.mjs:3:16)
      at tryCatch (node_modules/regenerator-runtime/runtime.js:62:40)
      at Generator.invoke [as _invoke] (node_modules/regenerator-runtime/runtime.js:288:22)
      at Generator.prototype.(anonymous function) [as next] (node_modules/regenerator-runtime/runtime.js:114:21)
      at asyncGeneratorStep (node_modules/@babel/runtime-corejs2/helpers/asyncToGenerator.js:5:24)
      at _next (node_modules/@babel/runtime-corejs2/helpers/asyncToGenerator.js:27:9)
      at node_modules/@babel/runtime-corejs2/helpers/asyncToGenerator.js:34:7
      at new F (node_modules/core-js/library/modules/_export.js:36:28)
      at node_modules/@babel/runtime-corejs2/helpers/asyncToGenerator.js:23:12
      at asyncToArray (src/async-to-array.mjs:19:24)
      at Object.asyncToArray (src/__tests__/take-sorted.test.js:30:18)
      at tryCatch (node_modules/regenerator-runtime/runtime.js:62:40)
      at Generator.invoke [as _invoke] (node_modules/regenerator-runtime/runtime.js:288:22)
      at Generator.prototype.(anonymous function) [as next] (node_modules/regenerator-runtime/runtime.js:114:21)
      at asyncGeneratorStep (node_modules/@babel/runtime-corejs2/helpers/asyncToGenerator.js:5:24)
      at _next (node_modules/@babel/runtime-corejs2/helpers/asyncToGenerator.js:27:9)
      at node_modules/@babel/runtime-corejs2/helpers/asyncToGenerator.js:34:7
      at new F (node_modules/core-js/library/modules/_export.js:36:28)
      at Object.<anonymous> (node_modules/@babel/runtime-corejs2/helpers/asyncToGenerator.js:23:12)

  ● asyncTakeSorted › return smallest iterable, using curry

    TypeError: Object is not async iterable

      1 | import { ensureAsyncIterable } from './internal/async-iterable'
      2 | 
    > 3 | export default async function asyncToArray (iterable) {
        |                ^
      4 |   const out = []
      5 |   for await (const item of ensureAsyncIterable(iterable)) {
      6 |     out.push(item)

      at _asyncIterator (node_modules/@babel/runtime-corejs2/helpers/asyncIterator.js:22:9)
      at _callee$ (src/async-to-array.mjs:3:16)
      at tryCatch (node_modules/regenerator-runtime/runtime.js:62:40)
      at Generator.invoke [as _invoke] (node_modules/regenerator-runtime/runtime.js:288:22)
      at Generator.prototype.(anonymous function) [as next] (node_modules/regenerator-runtime/runtime.js:114:21)
      at asyncGeneratorStep (node_modules/@babel/runtime-corejs2/helpers/asyncToGenerator.js:5:24)
      at _next (node_modules/@babel/runtime-corejs2/helpers/asyncToGenerator.js:27:9)
      at node_modules/@babel/runtime-corejs2/helpers/asyncToGenerator.js:34:7
      at new F (node_modules/core-js/library/modules/_export.js:36:28)
      at node_modules/@babel/runtime-corejs2/helpers/asyncToGenerator.js:23:12
      at asyncToArray (src/async-to-array.mjs:19:24)
      at Object.asyncToArray (src/__tests__/take-sorted.test.js:34:18)
      at tryCatch (node_modules/regenerator-runtime/runtime.js:62:40)
      at Generator.invoke [as _invoke] (node_modules/regenerator-runtime/runtime.js:288:22)
      at Generator.prototype.(anonymous function) [as next] (node_modules/regenerator-runtime/runtime.js:114:21)
      at asyncGeneratorStep (node_modules/@babel/runtime-corejs2/helpers/asyncToGenerator.js:5:24)
      at _next (node_modules/@babel/runtime-corejs2/helpers/asyncToGenerator.js:27:9)
      at node_modules/@babel/runtime-corejs2/helpers/asyncToGenerator.js:34:7
      at new F (node_modules/core-js/library/modules/_export.js:36:28)
      at Object.<anonymous> (node_modules/@babel/runtime-corejs2/helpers/asyncToGenerator.js:23:12)
```
If I comment out this in jest config:
```js
//    makeProject({
//      name: 'es5'
//    }),
```
it returns this instead:
```
 FAIL  src/internal/__tests__/iterable.test.js
  ● Test suite failed to run

    Cannot find module '../iterable' from 'iterable.test.js'

    However, Jest was able to find:
    	'../iterable.mjs'

    You might want to include a file extension in your import, or update your 'moduleFileExtensions', which is currently ['js', 'json', 'jsx', 'ts', 'tsx', 'node'].

    See https://jestjs.io/docs/en/configuration#modulefileextensions-array-string

      3 | const { range } = require('../..')
      4 | 
    > 5 | describe('ensureIterable', function () {
        |     ^
      6 |   it('works with iterables', function () {
      7 |     const i = range(3)
      8 |     expect(i).toBe(ensureIterable(i))

      at Resolver.resolveModule (node_modules/jest-resolve/build/index.js:202:17)
      at Object.<anonymous> (src/internal/__tests__/iterable.test.js:5:5)

 FAIL  benchmarks/async-map-parallel.js
  ● Test suite failed to run

    Cannot find module '../es2018/range' from 'async-map-parallel.js'

    > 1 | const range = require('../es2018/range')
        |                                          ^
      2 | const asyncMap = require('../es2018/async-map')
      3 | const asyncReduce = require('../es2018/async-reduce')
      4 | 

      at Resolver.resolveModule (node_modules/jest-resolve/build/index.js:202:17)
      at Object.<anonymous> (benchmarks/async-map-parallel.js:1:42)

 FAIL  src/internal/__tests__/querable-promise.test.js
  ● Test suite failed to run

    Cannot find module '../querable-promise' from 'querable-promise.test.js'

    However, Jest was able to find:
    	'../querable-promise.mjs'

    You might want to include a file extension in your import, or update your 'moduleFileExtensions', which is currently ['js', 'json', 'jsx', 'ts', 'tsx', 'node'].

    See https://jestjs.io/docs/en/configuration#modulefileextensions-array-string

      1 | /* eslint-env node, jest */
    > 2 | const querablePromise = require('../querable-promise')
        |                         ^
      3 | const delay = require('../delay')
      4 | 
      5 | describe('querablePromise', function () {

      at Resolver.resolveModule (node_modules/jest-resolve/build/index.js:202:17)
      at Object.<anonymous> (src/internal/__tests__/querable-promise.test.js:2:25)
```

This is only happening in this branch. Master works fine as usual

conartist6:
@sithmel The "object is not async iterable" failures are correct. You should be seeing those, because of the babel issue I mention. A fix for that has been merged into babel, we'll just have to wait for their next patch release. Also any sources we release with the intent of user transpilation will still need async generators pre-transpiled with that most up-to-date version of Babel for a while.

I believe the second set of results is a particular bit of stupidity that you can blame on jest. They have [an outstanding bug](https://github.com/facebook/jest/issues/7496) that if the projects array contains only one project it just doesn't work, the whole thing is ignored. What I have to do is copy the makeProject line, comment the whole projects block, then paste `...makeProject('es2015')`. Gross, but it works. Sorry you lost time to that.

sithmel:
@conartist6 I finally managed to play with it a bit. Thanks for the advice!

I think this is a useful feature, that comes basically for free. So as soon as everything is ready I am positive to merge (after this is ready, meaning all errors fixed, types, doc etc.).

The idea of iterators spawned on the fly is very smart indeed.

As for the implementation I would go back to the original class based implementation. Mostly because I found it more readable. I also suspect that it will produced a smaller transpiled version (no object.assign, to add as polyfill).

conartist6:
@sithmel Thanks! It's pretty cool that the reused currying set us up to do this, I never thought of it because it never would have been possible before. I'll finish writing everything up.

I do think that @KSXGitHub was right that the prototypal implementation will be smaller. `Object.assign` is a very small polyfill compared to what Babel needs to do for [classes with inheritance](https://babeljs.io/repl/#?babili=false&browsers=ie%2010%0A&build=&builtIns=false&spec=false&loose=false&code_lz=MYGwhgzhAEBCkFNoG8BQBfVrSRgMQHsDoEAPAFwQDsATGeCJNTbcKOMAJxIuro8YoMQA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=es2015%2Cenv&prettier=false&targets=Node-2.5&version=7.4.2). On the other hand the library we ship does not include that code, but rather depends on the babel-runtime module for those implementations, and it is possible/probable that in a normal environment, something else will depend on those same files anyway. That said it's not a complete given, as there are many possible semvers of babel-runtime, the loose mode flag would need to be off, and other package authors have to opt-in to taking the module as a dependency instead of just inlining the helpers.

sithmel:
> a very small polyfill compared to what Babel needs to do for classes with inheritance

wow, I didn't notice the size when you implement inheritance. Not a blocker anyway ...

conartist6:
OK should be merge-ready. Now I just have to go write `iter-tools-explode-imports`. Fortunately I was planning to do that anyway.

sithmel:
Working on the merge.

What is the "iter-tools-explode-imports" ?

conartist6:
I'm going to write a small babel plugin that does the following transform:
```js
import {map, filter} from "iter-tools"
//  becomes
import map from "iter-tools/es5/map"
import filter from "iter-tools/es5/filter"
```
I need this because I'm integrating with a project that has an old bundler which can't tree shake.

README.md
```diff
@@ -4,9 +4,34 @@ Iter-tools
...
+
+Iter-tools is at present the only fully-featured library of its kind. Here is what it supports:
+
+- **Isomorphic**: The tools run equally well in a browser and in node.js.
+
+- **Currying/Partial Application**: iter-tools methods can be passed their arguments all at once, or can be incrementally configured. E.g. either `map(x => ++x)(iterable)` is valid. Omitting the iterable argument allows methods to be pipelined easily, as the output of one simple transform will be suitable as the only unbound argument for the next. Iter-tools provides `pipe` and `execPipe` functions for this kind of chaining.
+
+- **null/undefined are empty iterables**: iter-tools methods treat both null and undefined as empty iterables. This allows you to write transforms on array-ish data without first needing to check whether the array exists.
+
+- **Multiple iteration**: iter-tools transforms (as of v7.0.0) can be evluated multiple times if the source iterator can be. Not all iterables can be evaluated multiple times though: generator functions, notably, cannot be. Arrays, Maps, and Sets (and null/undefined) always can be.
```
sithmel: The way you implemented it should work fine with generator functions (I tested it with range)

sithmel:
I left a comment and a PR to add on top of this to pass the test.
Apart that I am ready to merge this.

README.md
conartist6: Are you sure, because that doesn't make any sense. Range is tricky since it can never be exhausted, but you can never get it to go back to the beginning either. Any generator function that can be exhausted will just be stuck at the end, and attempting to iterate over it again will just result in an empty iteration as its first yield will be `{done: true}`.

sithmel: I tested it again, and I was wrong.

sithmel: Hey, I tried reimplementing range as:
```js
export default function range (opts) {
  return {
    [Symbol.iterator]: function * _range () {
      let start, step, end
      opts = typeof opts === 'number' ? { end: opts, start: 0 }
        : (typeof opts === 'object' ? opts : {})
      step = opts.step === undefined ? 1 : opts.step
      end = opts.end === undefined
        ? (step > 0 ? Infinity : -Infinity) : opts.end
      start = opts.start ? opts.start : 0

      for (let i = start; step > 0 ? i < end : i > end; i += step) {
        yield i
      }
    }
  }
}
```
And now it can be replayed as many times you need.
Combinatory functions already return iterables, so they works already

conartist6: Yep, great change.

PR #211, opened 2019-03-29
Add some names to .npmignore

KSXGitHub:


KSXGitHub:
So many commits, I should've pull first.

PR #215, opened 2019-04-18
remove tee/asyncTee, extra argument to fork, some linting fix

sithmel:
1 - Giving that the next release will be a major, I am getting rid of tee/asyncTee. Now fully replaced by fork
2 - I added an argument to fork to specify the number of forks. I think that came back a couple of times in the comments
3 - I am fixing some linting issues I have found in jest config files

Opening this as PR as I'd like to be sure we all agree about 1 and 2.


src/fork.mjs
```diff
@@ -48,7 +48,8 @@ export default function fork (iterable) {
   function * generateForks () {
     try {
       const consumer = exchange.getConsumer()
-      while (true) {
+      let counter = 0
+      while (counter++ < n) {
```
KSXGitHub: Why didn't you just use for-loop?

```javascript
for (let counter = 0; counter < n; counter++) {
  ...
```

sithmel: no reason. I should do that

sithmel:
This should be quite uncontroversial. I would go on merging

PR #220, opened 2019-04-22
Functions generating iterables can be reused

sithmel:


CHANGELOG.md
```diff
@@ -25,6 +25,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
  - takeSorted/asyncTakeSorted optional comparator is now the first argument
  - optional arguments can no longer have "undefined" as argument. This is at odds with considering undefined as a valid iterable.
  - asyncFlatMap takes an extra "concurrency" argument (default 1)
+ - range, cycle, asyncCycle, repeat, entries, values, keys, regexp-exec, regexp-split return iterables that can consumed multiple times
```
conartist6: Ah, I forgot to put a Changelog update in my PR. We should have an item about all the other functions being potentially consumable multiple times, but really there's no need to call out specific methods is there? With this change, all our iterables can now be consumed multiple times. I think that information belongs right up at the top of the changes section for 7.0.0

PR #221, opened 2019-04-23
Convert iterableCurry/asyncIterableCurry args to objects

conartist6:
The configuration objects need only be created once at parse time.
Doing this should save on both bundle size and execution time


KSXGitHub:
This change also improves readability, I like this.

PR #222, opened 2019-04-23
Generate sync/async implementations from templates

conartist6:
This is the proof of concept implementation which demonstrates that it is possible to use babel macros to maintain sync and async functions using a single set of sources.


conartist6:
I think the macro is now complete, which is to say it can handle all the syntaxes that vary between our sync and async code. I've added comments to it which would be helpful both to people who need to maintain the macro code and to people who just want to see at a glance what it does.

I am interested to hear if the consensus is still that this code simply adds too much complexity to be useful. My hope in showing an implementation is that it will be clear that such a solution may be approachable to contributors, and that its implementation would not be so overwhelmingly esoteric as to be unmaintainable by the other authors.

conartist6:
@sithmel I'm still looking to you for a go/no-go on continuing this work. I'm not going to do any more until I hear something.

sithmel:
For me is a no-go.
The solution you introduce has pros and cons.

The pros is removing repetition, but I don't see the amount of repetition we currently have, as an issue, not a huge one anyway.

The cons are:
- no correspondence between stack trace and source (when untranspiled)
- a barrier to contributors, but also occasional people debugging code

It seems to me that for removing a minor problem, you introduce a different (and arguably worse) problem.

sithmel:
Out of a conversation with @conartist6 : 
The two cons will be mitigated

> * no correspondence between stack trace and source (when untranspiled)

This should not be a problem as the transpiled code is part of the repo, so it should be transparent for debugging, and the comment will link to the template source.

> * a barrier to contributors, but also occasional people debugging code

Improving the readability of the template should mitigate the problem.

Ultimately, if in the future we have any regrets about this, it should be very easy reverting it back.

After these consideration I am happy to give a "go" on this.

conartist6:
So I added an `isAsync` export and renamed the macro (`a.macro` => `sync-to-async.macro`). I improved the documentation a bit further as well.

Finally jest is now configured so that the helpers can be used in test code (without creating duplicate results for tests which need only be run once). For now I am not using the template strategy for test files, but I could be convinced to alter the approach I've taken. Certainly as I found when debugging it can be tricky if the plugin introduces a syntax error (which it never should once this PR is fully developed) to figure out what the test file looked like at parse time.

We had discussed possibly using the sync file as the template for the async file, but I realize that would negate other advantages as the sync file would then not have the same source before and after macro application.

conartist6:
Let me know what other readability improvements you can think of. One option I considered is creating even more symbols. They could be called, perhaps, `asyncable` and `mayWait`, and would be used only in the spots that their respective keywords would be applied.

src/__tests__/a-reduce.test.js
```diff
@@ -0,0 +1,59 @@
+/* eslint-env jest */
+import a, { isAsync } from '../macros/sync-to-async.macro'
```
sithmel: I would prefer using a more descriptive name instead of "a". Possibly containing the word "macro"
"syncToAsyncMacro" ?

conartist6:
How about this:
```js
import { isAsync, iteratorSymbol, $async, $await } from "sync-to-async.macro"
```

conartist6:
We could change the other two as well to `$isAsync` and `$iteratorSymbol` such that $ would be a quick visual cue to the reader that macro transpilation was in play.

conartist6:
I rather like that idea. It allows us to be as close as possible to the keywords that we can't use. It has minimal overhead in terms of taking up screen real estate, and it gives us an easy tip for new users that we can put right up front in CONTRIBUTING.md, something like:

> If you are looking at code and see a variable name beginning with a $, the variable is defined by a macro, and is used to mark a difference between the implementations of the synchronous and asynchronous functions.

conartist6:
OK I think it's looking pretty close to how it should.

TODO:
- Add the script that ensures that running generate doesn't create any changes under version control (or untracked files) and ensure that Travis runs it.
- Ensure that generate script removes generated files when their source templates are removed

sithmel:
$async, $await is much better!

conartist6:
Remaining TODO:
Update the readme with documentation for the new parallel functions. These now have a default concurrency. I chose 4 because it is [guaranteed to be fair and random](https://xkcd.com/221/).

conartist6:
This now depends on https://github.com/kentcdodds/babel-plugin-macros/pull/113

conartist6:
The problem of how to ensure that removed templates don't leave their generated files in the repo is a bit tricky. It's easy enough to handle in watch mode, but it's much harder otherwise. I thought about using git to find out if any template files had been deleted, but if the bad code has already been committed then the tooling will hum along happily.

The strategy I'm going to pursue is to create a jsdoc-style comment attribute, and a new eslint plugin. Each file will start with something like
```js
/**
 * @generated-from ./some.template.js
 * ...
 */
```

The linter will pick up on the @ comment syntax, and will try to stat the template file. If it does not exist, this will be a lint error.

This should work great since husky already runs the linter as pre-commit step. Combined with the watcher I think this change should add relatively minimal additional overhead for people like @KSXGitHub who don't have the best hardware.

conartist6:
To make the linter happy with on the generated files we'll also need https://github.com/babel/minify/pull/953

conartist6:
To fix: better test coverage shows that parallel methods return values out of order at higher levels of parallelism

conartist6:
The issue is that `previousUnresolvedPromise` is not sufficient for parallelism higher than two. At higher levels of parallelism we need to track multiple other inflight promises to ensure that we do not emit out of order.

conartist6:
I'm thinking that when the generate script errors it should replace the generated script with:
```js
/**
 * @generate-failed
 */
throw new Error(${JSON.stringify(codegenError)}")
```

Since I'm planning to use the linter in combination with `comment-parser` to pick up on jsDoc-like comments it would be easy enough to make the `@generate-failed` tag a lint error, thus preventing these files from actually ever making it into the upstream repo.

conartist6:
This code generator, if you can't tell, is not going to suck. I've had too much of my time wasted by mediocre code generators to write another one.

conartist6:
Here's how it looks to jest when the codegen step fails:
Jest error

sithmel:
Out of curiosity:
why do you stringify a string in ```${JSON.stringify(codegenError.toString())``` ?

Also, why don't add something to the message to identify that the error is coming from the codegen step ?

Something like "codegen: ...rest of the message ..."

conartist6:
Yeah I can definitely add something to the error to call out that it's from the codegen.
I stringify the string because it's not safe to just interpolate it. It probably contains quotes!

conartist6:
My PR to babel-plugin-macros was accepted and shipped. I hope to have a final version of this PR tonight.

One interesting thing to note is that apparently the attributes file I added to cause github to ignore generated files actually activated the git-lfs (large file storage) plugin, which uses a different underlying storage mechanism. In theory this is good, it means that the git core itself will be performing appropriate optimizations. In practice it's another thing that everyone needs to install (though evidently my personal laptop had it set up already?). Let me know if you have any concerns around this. The solution does not strictly speaking depend on it, though I very much like the idea of removing the generated files not needing explicit human review.

conartist6:
Go flight.

README.md
```diff
@@ -407,11 +421,18 @@ flatMap(x => [x, x * x], range(4)); // 0, 0, 1, 1, 2, 4, 3, 9
 
 ## async-flat-map
 Same as flatMap but works on both sync and async iterables.
-It can also use as extra argument, the concurrency level (default 1):
+
+## async-map-parallel
```
sithmel: should be async-flat-map-parallel

conartist6: Fixed

conartist6:
@sithmel I'm on gitter too if you're going through this now and have any questions

sithmel:
I didn't run through the whole PR. I would take a day :-)

But there is everything I expect (and much more). Great job!

Feel free to merge @conartist6

conartist6:
That's great, thanks! I'm sure some things will come up, but we're not releasing 7.0 immediately and I've put quite a lot of polish work into this already, so I'm going to go ahead and merge.

PR #223, opened 2019-04-30
Fix asyncBatch type definition

wong2:
The result should be an array


sithmel:
Sync version is already fine. Merging

sithmel:
Thanks a lot @wong2

PR #225, opened 2019-05-13
Improve flat API

conartist6:
As I am working my way through the library eliminating async duplication, and I came across our implementation of flat, which I believe has some problems:

It allows you to specify depth as a number, unless of course it doesn't. If you want to pass a custom flattening predicate, you can no longer control depth using a number, but are in essence expected to implement that part of the function's logic yourself.

It also has a particularly nasty case: `[...flat("abc")]`. The current result is `["abcde"]` but the correct result should be `["a", "b", "c"]`, which is the same as the result of simply `[..."abc"]`.

### DIFF
--- a/src/async-flat.mjs
+++ b/src/async-flat.mjs
@@ -2,7 +2,12 @@ import { asyncIterableCurry } from './internal/async-iterable'
 
 const defaultShouldFlat = item => (typeof item[Symbol.iterator] === 'function' || typeof item[Symbol.asyncIterator] === 'function') && typeof item !== 'string'
 
-function flat (shouldFlat = defaultShouldFlat, depth = 1, iterable) {
+function flat (shouldFlat = defaultShouldFlat, depthOrOptions = 1, iterable) {
+  let depth = depthOrOptions
+  if (depthOrOptions && typeof depthOrOptions === 'object') {
+    ({ shouldFlat = defaultShouldFlat, depth = 1 } = depthOrOptions)
+  }
+
   async function * _flat (currentDepth, iterable) {
     for await (const item of iterable) {
       if (currentDepth < depth && (await shouldFlat(item))) {
--- a/src/async-flat.mjs
+++ b/src/async-flat.mjs
@@ -1,31 +1,19 @@
 import { asyncIterableCurry } from './internal/async-iterable'
 
-const defaultShouldIFlat = (depth) => {
-  if (typeof depth === 'function') {
-    return depth
-  }
-  if (typeof depth === 'number') {
-    return (currentDepth, iter) =>
-      currentDepth <= depth &&
-      (typeof iter[Symbol.iterator] === 'function' || typeof iter[Symbol.asyncIterator] === 'function') &&
-      typeof iter !== 'string'
-  }
-  throw new Error('async-flat: "depth" can be a function or a number')
-}
+const defaultShouldFlat = item => (typeof item[Symbol.iterator] === 'function' || typeof item[Symbol.asyncIterator] === 'function') && typeof item !== 'string'
 
-function asyncFlat (shouldIFlat = 1, iterable) {
-  shouldIFlat = defaultShouldIFlat(shouldIFlat)
-  async function * _asyncFlat (currentDepth, iterable) {
-    if (await shouldIFlat(currentDepth, iterable)) {
-      for await (const iter of iterable) {
-        yield * _asyncFlat(currentDepth + 1, iter)
+function flat (shouldFlat = defaultShouldFlat, depth = 1, iterable) {
+  async function * _flat (currentDepth, iterable) {
+    for await (const item of iterable) {
+      if (currentDepth < depth && (await shouldFlat(item))) {
+        yield * _flat(currentDepth + 1, item)
+      } else {
+        yield item
       }
-    } else {
-      yield iterable
     }
   }
 
-  return _asyncFlat(0, iterable)
+  return _flat(0, iterable)
 }
 
-export default asyncIterableCurry(asyncFlat, { variadic: false, minArgs: 0, maxArgs: 1 })
+export default asyncIterableCurry(flat, { minArgs: 0, maxArgs: 2 })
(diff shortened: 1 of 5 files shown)


conartist6:
@KSXGitHub I'm also getting an interesting message from typescript in your spec file:

```ts
// flat(0, iterable)
assert<{
  'depth = 0': IterableIterator<0 | 1 | 2>
  // ...
}>({
  'depth = 0': iter.flat(0, [0, 1, 2] as [0, 1, 2]), // Type instantiation is excessively deep and possibly infinite
  // ...
})
```

It doesn't fail the tsc check though.

KSXGitHub:
It was the new changes in TypeScript 3.4 that break one of the dependencies. I tried to fix it once but I gave up.

conartist6:
I also think that for the readme we should adopt some standard method of defining a function API. e.g. flat could be written as:
`flat([[shouldFlat], depth], iterable)`

Of course that argument syntax is ambiguous in javascript even more than it is in bash, but it has the advantage of conveying quite clearly something which is tricky to describe succinctly and clearly in english: the order that optional arguments appear in.

conartist6:
@KSXGitHub which dependency?

KSXGitHub:
> @KSXGitHub which dependency?

typescript-tuple

src/async-flat.mjs
```diff
@@ -1,31 +1,19 @@
...
-  }
-  if (typeof depth === 'number') {
-    return (currentDepth, iter) =>
-      currentDepth <= depth &&
-      (typeof iter[Symbol.iterator] === 'function' || typeof iter[Symbol.asyncIterator] === 'function') &&
-      typeof iter !== 'string'
-  }
-  throw new Error('async-flat: "depth" can be a function or a number')
-}
+const defaultShouldFlat = item => (typeof item[Symbol.iterator] === 'function' || typeof item[Symbol.asyncIterator] === 'function') && typeof item !== 'string'
```
KSXGitHub: you may want to split this function into multiple lines

conartist6: The reason I didn't is that once we are generating from templates we'll lose control of formatting, and the next changes to this file will likely be generating it from a template. Given that that's likely to be the next change to this file, I didn't want to cause unnecessary formatting churn.

If we want to maintain line-length-limited formatting for our generated code we'll need to adopt prettier, which is a whole other discussion.

KSXGitHub: I see

sithmel:
> It allows you to specify depth as a number, unless of course it doesn't. If you want to pass a custom flattening predicate, you can no longer control depth using a number, but are in essence expected to implement that part of the function's logic yourself.

Yes, why do you consider this a problem ?
I found it more flexible, you can decide to use different depths depending on what you are flattening
Having 2 arguments has the additional problem of making the function call more difficult to remember. What is the first argument? is it the depth or the function ?

> It also has a particularly nasty case: [...flat("abc")]. The current result is ["abcde"] but the correct result should be ["a", "b", "c"], which is the same as the result of simply [..."abc"].

Your fix makes sense

conartist6:
APIs for controlling the behavior of a function shouldn't appear or disappear depending on what other options you pass unless the APIs are actually incompatible with each other.

You could perhaps make a case that depth should still be a second argument to `shouldFlat`. I see neither any major drawbacks to exposing it that way, nor any significant use cases that would require it.

conartist6:
Also in general I thought our solution to the problem of remembering how arguments go was that as a rule we would try to append optional arguments to the beginning of arguments lists. One problem with that is that I would dearly like for slice to more or less match its public API, and its optional arguments are appended: e.g. `slice(1)` and `slice(1, -1)`. If we're prepending arguments that second call should really look like `slice(-1, 1)` but I don't think that's what anyone is expecting to see. That's part of why I was discussing some kind of format for the docs that can make this explicit, which I think could look like either of these examples:
```js
flat(depth, iterable)
flat(shouldFlat, depth, iterable)
```
or
```
flat([[shouldFlat], depth], iterable)
```

conartist6:
As a note the reduce builtin also appends its optional arguments, and we currently prepend them.

KSXGitHub:
I don't see the need for `depth` when `shouldFlat` is present. User can easily use closure for their own `depth`.

So the code you wrote:

```javascript
flat(shouldFlat, depth, iterable)
```

can be turned into:

```javascript
flat((...args) => shouldFlat(depth, ...args), iterable)
```

conartist6:
`depth` is the `shouldFlat` callback isn't the specified depth to flatten to, it's the current level of recursion, e.g. 1 if it's inside one layer of iterables, 2 if it's inside two layers, etc.

KSXGitHub:
> `depth` in the `shouldFlat` callback isn't the specified depth to flatten to

To clarify, I have edited my comment.

Anyway, why do `depth` and `shouldFlat` both present? Aren't they both decide when to "dive deeper"?

conartist6:
As I said in my earlier comments, APIs should not just disappear unless there is an incompatibility.

Especially since we are emulating a builtin, `Array.prototype.flat`, which can only ever take the depth argument. We should be adding to the basic API, not replacing it with something which is meant to accomplish the same goal but works a completely different way that we force you to do some of the time.

sithmel:
If we are going to have depth and shouldFlat as separated arguments, can we have them passed as a single "options" argument ?
This is exactly what we do in "slice".
case 1:
```js
flat(iterable)
```
case 2:
```js
flat(depth, iterable)
```
case 3:
```js
flat({ depth, shouldFlat }, iterable)
```

My concern is that having:
```js
flat(depth, shouldFlat, iterable)
```
(both optional arguments) can be difficult to remember

conartist6:
I think we should allow the optional arguments, but I don't mind also allowing named options. I think two optional arguments is OK.

sithmel:
If we have the option of using named arguments is ok for me.

conartist6:
Added the option of named arguments, though I have omitted typescript support for the named arguments. My reasoning is that typescript code will be developed in an environment which can already tell you the names of positional arguments. Also given the change I suggest [here](https://github.com/sithmel/iter-tools/issues/226#issuecomment-492712466) adding the additional overloads would result in a nightmarishly large type.

PR #235, opened 2019-06-16
Rewrite $ identifiers

conartist6:


conartist6:
In essence I originally hoped to be able to shift all work to the macro, but names of things were the hardest. The worst thing about it is that all the async method names were wrong, so in a stack trace `asyncMap` would show up as `map`.

This change fixes that, eliminates the need for additional infrastructure to allow tests to import sync or async variants easily using $, and also is what we need in order to use our system on our typescript definitions.

PR #253, opened 2019-07-12
Use templating system on typescript types.

### LINKED ISSUE #237: Port typescript types to templates
Just creating this to track the work that I'm doing.

Now that we have implementation templates for methods, we should have type definition templates too since our type definitions are similarly copied and pasted between our sync and async methods.

This means first that the import structure of type definitions will be inverted to match the source import structure, something I have long wanted to do anyways. The actual type definition for `src/batch.js` will now live in `src/batch.d.ts`, while `index.d.ts` will simply import and reexport the various method definitions.

This can then be templated, such that `src/$batch.d.ts` lives next to `src/$batch.js` and generates both its sync and async typedefs.

conartist6:
Fixes #237.

There's a lot going on here. Some of it includes:

Types are no longer specified in one huge file. The types for a method are specified in a file alongside the method.

Prettier is now used to format all files in the repo. The code generator formats with prettier prior to output now.

Changes were made to the code generation system. The generate command (`npm run generate`) now triggers multiple generators. This may currently lead to inefficient use of file handles with the default watcher, so it is recommended for the moment to pass the new `--watchman` parameter.

New type for the `flat` method.

Numerous bugs in async types fixed.

TODO:
Ensure files with incorrect prettier formatting cannot be checked in.


conartist6:
By the way, I don't recommend trying to read the PR. It's just too complicated. Better to just check out the code and poke around.

conartist6:
Hey, where did Travis go?

conartist6:
Looks like I’m not an owner of the organization so I can’t get Travis back. @sithmel you should have gotten an email from Travis, and/or you could just make me an org owner.

sithmel:
@conartist6 You are the owner now

conartist6:
There we go. Dunno why Travis didn’t pick it up before. Anyway, thanks!

KSXGitHub:
I don't think we should add generated files to git.

If I was you, I would create a folder called `generated-types/` which contains one `.keep` in git tree and multiple generated `.d.ts` files. `src/index.d.ts` would import/re-export things from `generated-types/*.d.ts`.

conartist6:
We made an intentional choice that it was worth checking generated files in to git so that the history of the method implementations in plain javascript (not a templating pseudo-language) would always be viewable in the repo’s git history. This seemed particularly important to me because templating language we’re using is not likely to spread beyond this project, and will probably continue to adapt to our code generation needs. People trying to see the history of the code should not also need to consider the history of the templating system at the same time.

KSXGitHub:
@conartist6 Okay then. I would like every generated file to have comment at the start clarifying that "this is a generated file"

conartist6:
They… all… do?

KSXGitHub:
Okay. I double-checked. They all do.

conartist6:
Also Travis checks that running the code generator does not result in any git changes or untracked files, which prevents people from accidentally pushing stale templates to master.

conartist6:
Made the linter run on the generate directory and cleaned up lint errors.

KSXGitHub:
> Made the linter run on the generate directory and cleaned up lint errors.

I'm confused. Is this supposed to be a commit message?

conartist6:
I force pushed. I would consider splitting this up into multiple commits, but so far I haven’t.

PR #256, opened 2019-07-29
Add when method (take 2)

conartist6:
I still really want a method that helps do conditionals with the spread operator. I proposed something a while ago on this topic, but it was a rather janky and complicated API that was out of place in iter-tools. I believe that I've now simplified the method considerably.


README.md
```diff
@@ -977,7 +978,41 @@ Some of these functions makes an in memory copy of the output. For example: cycl
...
+  )
+} // [true, true] OR [true]
+```
+
+Instead, you can use the when method like so:
+```js
+const whenArr = {
+  always,
+    ...when(sometimes, [ sometimes ]),
+  ) // [true, true] OR [true]
```
KSXGitHub: I think you should remove this closing brace

src/__spec__/when.spec.ts
```diff
@@ -0,0 +1,16 @@
+import assert from 'static-type-assert';
+import { when } from '..';
+
+declare var Ø: never;
```
KSXGitHub: Fancy, but I would prefer English.

KSXGitHub:
I don't think this is very useful.

Even if this is merged, I would still use ternary because:
  * Ternary requires no function definition, and thus produces a lighter code.
  * Code reader have to look up definition of `when` to know its use, and then probably be confused about why such function is necessary. But ternary is obvious.

src/__spec__/when.spec.ts
conartist6: I couldn’t think of a good name for something that doesn’t exist. I wanted a really short name because this is used across the type specs now. I thought of a names like `any` or `never`, but those are the names of types, and would easily be confused with the actual types when trying to read the code. I liked Ø because it’s short, can mean “the empty set” in mathematical notation, and is visually distinct, which draws attention to positional parameters, sort of like a comma.

README.md
conartist6: And the outer braces should be `{` not `[`.

src/__spec__/when.spec.ts
KSXGitHub: Why didn't you do this?

```typescript
declare const ANY: any;

assert<Iterable<number>>(when(ANY, [] as Iterable<number>));

assert<[1, 2, 3] | []>(when(ANY, [1, 2, 3] as [1, 2, 3]));

assert<Object>(when(ANY, ANY as Object));

assert<{test: 'test'} | {}>(when(ANY, {test: 'test' as 'test'}));

assert<{}>({...when(ANY, null)});

assert<Array<never>>([...when(ANY, null)]);
```

conartist6:
Yeah I agree that in many cases it still makes sense to use ternary just because then there’s no need to look anything up. However there are certain types of code in which this pattern is quite common, and thus the improvements to formatting and thus readability would be significant. I  also think the function implies its workings reasonably well using the syntax boilerplate around it, usually something like `[first, …when(hasMore, [ second, third ]]`. In my mind there’s no need to look that up.

src/__spec__/when.spec.ts
conartist6: My pattern seemed simpler? There’s only one thing that occurs in a value position, and all type annotations are completely explicit, which is exactly what I want in a type test.

sithmel:
> Code reader have to look up definition of when to know its use, and then probably be confused about why such function is necessary. Ternary, in the other hand, is obvious.

I agree with @KSXGitHub about this.

Also let's assume this has a use case; why this should be in iter-tools ?

conartist6:
Why shouldn’t it be in iter-tools? Ignore for a second the fact that it also works for objects. It’s a function defined over iterables, just like `map` or any other method in this library. Obviously you can always map an iterable with `[…iterable].map()` and many people do. Many people will continue to do that just because the perf doesn’t matter to them and it’s easier not to add an external dependency.

However, it adds one of the basic iter-tools value propositions to the pattern: null is treated as an empty iterable. If you have bought into our philosophy of allowing null to represent an empty iterable, the plain js pattern you’d need would actually be `[…(value != null && condition ? value : [])]`. Note that you need to include value twice on that line, which is a pain if value is a long expression. Faced with writing a more complicated, non-DRY expression just for a bit of null safety, which may not even be necessary (yet), most people would choose the shorter expression, and could be burned later if the ability to use null as an empty iterable ever was needed.

conartist6:
Er, note that null wasn’t being treated correctly. Pushing that fix.

conartist6:
Note also that while the implementation is trivial to us, there were a few iterations of refining the API, and a lot of the value is in the documentation, tests, and types. I also think that the fact that the function can work the same way for objects adds value, and the mechanisms allowing that to work from both a runtime and type perspective are not immediately obvious.

src/__spec__/when.spec.ts
conartist6: Also I gravitate towards rules that I can explain as simply as possible, given that I hope that eventually this project will no longer require a significant amount of my time to maintain. I could easily write a lint rule that says in spec files only `Ø` should go in the value position. That would guarantee that it is the responsibility of the `__spec__/*.spec.ts` files to test type evaluation only. That way no-one is tempted to do things like use `1` as a shorthand for `number`, not be confused with `1 as 1`.

PR #258, opened 2019-07-30
Add firstOr and notFound arg for find. Changelog fixes.

### LINKED ISSUE #229: Give first a default parameter to be returned if iterable is size 0
Idea lifted from https://github.com/ReactiveX/IxJS/issues/275

conartist6:
Fixes #229


conartist6:
The general guiding principle here is that it should be possible to disambiguate between the first/found item existing and having the value `undefined` and there being no first/found item.

PR #266, opened 2019-08-07
Add wrap method. Range positional args and zipAll filler arg.

conartist6:


conartist6:
This also introduces a new internal function, `wrapWithMethodIterable` which is designed to replace this pattern:
```js
function someMethod() {
  return {
    [Symbol.iterator]() {
      // …
    }
  }
}
```
The problem with that pattern is that the result is a reusable iterable, but it is not also an iterator, which the rest of our results are.

conartist6:
Sorry this branch might keep getting a little bigger. It’s a lot of work to handle all the merges and sequencing of PRs, so I might lump a few things together while they’re await feedback. I’m keeping the commits distinct though.

PR #270, opened 2019-08-20
String inspired methods

conartist6:
This is a big one. Basically the idea was to extract iter-tools methods from string methods. After all strings are iterables, and now the iter-tools wins on handling `null` and `undefined` should be applicable to the common string operations. I've ensured that the methods have performant implementations when applied to regular strings.

I highly recommend beginning your review with the CHANGELOG. I suspect more of the discussion will be around what is in that file than anywhere else.

You'll note that I've introduced the idea of "method families". This is in response to the need to disambiguate with some of these methods. Lets take `includes` as an example:
```js
includes([1, 2], [[1, 2], 3]])
includes([1, 2], [1, 2, 3])
```
Which of these is true?
I've opted to call the second of them `includesSubseq`. I had initially considered a config object like `{subseq: true}`, but there were several things I disliked about it. One, the parameter changed the fundamental behavior of the method, including its type signature. Two, it forced you to write out `includes({ subseq: true }, seq, it)` which is 12 characters longer than `includesSubseq(seq, it)`. Three, the `true` literal distracts from the natural sentence-like structure of the call.

I then add a second dimension with `any`. This was most directly inspired by the needs of `splitLines` (now called `linesFromChunks`), but lets look at `includes` again for simplicity's sake. When one is iterating over a file or similar consumable-only-once iterable, you may need to check that one of several things is included. Lacking an `includesAny` implementation, you'd be forced to iterate multiple times, one iteration for each possible included element. This means that you'd have to cache the entire iterable, which is a needless cost. For methods like `splitOn` the iterable could even be infinite, precluding a caching solution.

I feel it is acceptable to have these two dimensions meet cross-product style, because I anticipate that any further configurability in these methods will not change their type signature/fundamental behavior, and thus would be expressable as an optional configuration parameter. My idea is that by calling this a "family of functions" we can reduce the clutter associated with it in the docs, e.g. by referring to the family instead of all of its members, and by creating UI in the website that presents the functions together.

Finally, the state of the code I've posted: Testing is lagging behind the implementations at the moment, in part because I want to see what feedback I get here before I do polish work. I intend to add `equals` `equalsAny`, `deepEquals`, and `deepEqualsAny` methods in a follow-on PR.


conartist6:
Ah, another note: as a convenience, strings always work with both subseq and non-subseq methods since there is no ambiguity with them. For example:
```js
includes('abc', 'bc'); // true
includesSubseq('abc', 'bc'); // true
```

KSXGitHub:
I disagree with "strings work with non-subseq":

1. It is inconsistent with other types of iterable.
2. I don't want to pay extra cost (in file size and runtime check) for non-string iterables.
3. `String.prototype.includes` is good enough, I don't need a library sich as iter-tools.

conartist6:
Ah, but `String.prototype.includes` doesn't work on `null`. And you're not really paying much extra cost, because I am just delegating to a built in string method in the case where I get a string.

conartist6:
I do appreciate the argument that it's inconsistent though. I am striving for consistency. Here I chose the balance on ease of use because in the actual code it was relatively easy to implement.

KSXGitHub:
> but `String.prototype.includes` doesn't work on `null`

[Optional chaining](https://github.com/tc39/proposal-optional-chaining) does: `null?.includes()`

conartist6:
Ah that's a good point. Anyway the method will work on null, and it's definitely going to work on iterables including strings. And I don't see any problem with delegating to native code where such code exists.

conartist6:
Big update here. Everything is rebased. `equal` method is in. `deepEqual` will can come later. Removed `asyncSplitChunkedOn` which was previously here. I'm thinking of making a separate library just for fs iter tools, e.g. open a file as a line iterator, or open a directory as a file path iterator.

conartist6:
For some reason the checks don't want to run on travis. `npm run verify` passes locally though.

PR #274, opened 2019-08-26
Reorganize

conartist6:
I currently plan to merge this ahead of #270 and rebase.

I was inspired by the difficulty of working with any of the large directories or files now that there are so many methods. So I'm going to break things down better!


conartist6:
Here's some of the other changes that will be coming to this PR

Make the code generator take care of creating files (implementations and typedefs) in `src/` that point into `methods/`.

Explode `README.md` into `src/methods/*/README.md` files.

Create a code generator that builds an API.md from `README.md` files.

Root readme should reference API.md

conartist6:
Finally just browsing [the file tree](https://github.com/iter-tools/iter-tools/tree/reorg/src/methods) is enough to understand at a glance what methods are part of the library, including which are sync/async/both.

conartist6:
One more reason for what I'm doing here.

The contents of the `methods/` folder should never be imported from outside the library, and thus can fully leverage the es6 import/export syntax. This allows us to make a separate export of the uncurried method for use internally to other methods, where we don't want all the currying and argument validation logic to have to execute again.

conartist6:
Seems like maybe rename detection didn't really work in the PR... TODO: investigate.

PR #280, opened 2019-09-02
Improve combinatorics implementations

### LINKED ISSUE #275: Factorial cache of unbounded size constitutes a memory leak

### LINKED ISSUE #276: Permutations implementation is not performant
It does far more steps than it might need to, e.g. for `permutations(1, range(100))` it should only need to do 100 steps, yet it actually does 100 factorial steps! And it creates a `Set` at every single step. So yeah, I rewrote it. Also I forgot what a pain in the ass it is to flatten recursive algorithms. Spent so much time just debugging... Anyway. 

https://gist.github.com/conartist6/da1ec9ba4bf2d506ce7a4818f3fc503d

conartist6:
Fixes #276
Fixes #275


src/methods/combinations-with-replacement/combinations-with-replacement.mjs
```diff
@@ -16,20 +16,19 @@ function isSorted(arr) {
 }
 
 export function combinationsWithReplacement(iterable, r) {
-  const arr = Array.from(iterable);
+  const arr = [...iterable];
```
KSXGitHub: Why do you use spread operator? `Array.from` was chosen because babel does not fill resulting code with helper functions.

src/methods/permutations/permutations.mjs
```diff
@@ -1,26 +1,55 @@
...
+          ist.pop();
+          start--;
+          ist[start]++;
         }
       }
     },
-    getSize() {
-      return permutationsSize(len, r);
+    get size() {
+      return permutationsSize(arr.length, k);
```
KSXGitHub: Function (`iterable.getSize()`) was because it contains computation and thus must be explicit in use: User must call the function, and therefore, when using it twice, they opt to call the function once and assign its result to a variable.

Property access (`iterable.size`) does not discourage user from calling it twice as they assume data access contains little to no computation, resulting in worse performance.

src/methods/product/__tests__/product.test.js
```diff
@@ -21,19 +21,48 @@ describe('product', () => {
...
+      [4, 3],
+      [4, 4],
+    ]);
+  });
+
+  it('works with different length inputs', () => {
+    const iter = product([1, 2], [3, 4, 5]);
+    expect(iter.size).toEqual(6);
+    expect(Array.from(iter)).toEqual([[1, 3], [1, 4], [1, 5], [2, 3], [2, 4], [2, 5]]);
+  });
```
KSXGitHub: You have two assertions in one `it`.

I would do this instead:

```javascript
  describe('works with different length inputs', () => {
    const iter = product([1, 2], [3, 4, 5]);
    
    it('size property', () => {
      expect(iter.size).toEqual(6);
    })

    it('contains expected elements', () => {
      expect(Array.from(iter)).toEqual([[1, 3], [1, 4], [1, 5], [2, 3], [2, 4], [2, 5]]);
    })
  });
```

src/methods/permutations/permutations.mjs
conartist6: I'd rather do one of two things: do the computation up front even before the user attempts to access `size`, or just cache size for them after the first call to the getter. The second approach promises to be fastest.

src/methods/combinations-with-replacement/combinations-with-replacement.mjs
conartist6: Ah, I hadn't been thinking of that. I would have hoped that babel would just transpile it to `Array.from`. I just put it that way because in the long run I think language features are preferable to builtins.

src/methods/permutations/permutations.mjs
KSXGitHub: > do the computation up front even before the user attempts to access `size`, or just cache size for them after the first call to the getter

Implementation of `get size` does not appear to do either of these things.

src/methods/product/__tests__/product.test.js
conartist6: I didn't create this pattern here. And I really don't think there's anything necessarily wrong with having two assertions in one `it`.

KSXGitHub: > I really don't think there's anything necessarily wrong with having two assertions in one `it`.

When it both assertions fail, you can only see the first error and not the second.

conartist6: I feel confident saying that I'm not worried about that problem since I screwed both parts of those assertions up several times while doing this refactor, and it was only the smallest of inconveniences that I didn't get both errors at once.

src/methods/permutations/permutations.mjs
conartist6: I have changed it to cache size.

src/methods/combinations-with-replacement/combinations-with-replacement.mjs
conartist6: Actually to be honest I'm almost certain I prefer the babel helper here. `Array.from` is tricky. Some platforms/browsers implemented `Symbol.iterator` before they implemented `Array.from`. Others implemented them in the reverse order, meaning some versions of `Array.from` in the wild *may not actually support iterables*. Contrast that with the babel helper, which I am certain always does the same thing. And the helper is a module, so there should only be one copy of it across the entire project. Given that we use array spread in other code, it's already going to be part of the bundle anyway.

conartist6: And I believe that was the last `Array.from` outside test code, so now the library no longer needs it to exist/be implemented correctly at all to work.

PR #281, opened 2019-09-03
Add yeoman and `create:method` script.

conartist6:
The manual code generator is designed to reduce the amount of typing and copy/pasting even more.

It collects several pieces of information, each of which can be specified on the command line or is collected via interactive prompt if it is not specified. They are: the name of the new method, whether the new method is reducing, and whether the implementation is sync, async, or both($).
Full validation is done to ensure that options cannot be specified in an invalid or inconsistent manner.

Here's what it looks like using the tool:
Screen Shot 2019-09-03 at 9 35 59 AM


conartist6:
As a note, it would be significantly easier to use our internal scripts for people who have yarn. Specifically npm has some unpleasant behavior around passing arguments to scripts. For example in this branch, the following will work:
```
npm run create:method foo
```
The script gets the argument `foo`. However the same is not true of a flag:
```
npm run create:method foo --sync
```
After running the above command, the user will not be prompted for the method name, but will be prompted for whether or not the method should be sync, because in fact `--sync` has been interpreted as an argument to `npm run`. Yarn avoids this silliness by requiring that any args to the script runner appear before the name of the script. Anything after the name of the script is passed to the script itself, which is almost always what is expected.

conartist6:
Also this means that running `method:create` and `generate` is enough to get you a whole live working method implementation from nothing. All the other boilerplate (like adding the method to `index.mjs` and `index.d.ts` is automatic now. The method will even automatically be added to the documentation, just as soon as I finish building the code to merge the individual method READMEs into one `API.md`.

KSXGitHub:
I see you are adding more and more devDependecies as time went on, and the project's complexity increases ever higher. When such case arises, I usually turn my repo into a monorepo, with every dev tools in one folder separated from the npm package itself. Doing so would also make splitting the package into smaller ones easy: For example, sometimes I only need `concat` function so I only install `@iter-tools/concat`. I have [such a repo](https://github.com/tsfun/tsfun), you may adapt it to your own need. (If you care about this request, I may move this to a separate issue)

conartist6:
The biggest thing that I plan to split out is the current code generator living in `generate/generator` to its own package. Then `generate` and `generate-yo` will be similar, witch each just definition configurations which get plugged into an imported framework. That way the only things which will stay in this repository are those that are directly related to how iter-tools is organized and built.

I definitely do not intend to make iter-tools a monorepo. It would not solve any problem, and would complicate development and consumption of the library considerably.

conartist6:
In the situation you mention with `concat`, your suggestion is important when you are in a node project in which `node_modules` size is a critical concern and you don't bundle your code and you only need a very few methods.

That is just too small a use case to structure the whole library around. You have many options if you are in the above situations, including writing your own `concat` method (trivial, with generators), or better bundling with rollup prior to shipping your space-sensitive compiled app, e.g. as an Electron app.

conartist6:
I should also say that it would be easy to publish the yo generator as its own package. In fact yo generators are really meant to be their own packages. I specifically wanted to ensure that the templates and the library stay in sync, because the basic structure of a method can change even when the library's semver doesn't. I couldn't think of a good way to ensure that an external generator package would always make the correct boilerplate for the version of iter-tools being developed against.

PR #289, opened 2019-09-15
API.md generator and DOCME files. Move parallel methods.

conartist6:
This is the last major stage of the effort to use code generation to automate manual maintenance. We now generate the API docs from the documentation for individual methods.


conartist6:
I recommend reviewers begin with a look at the output: https://github.com/iter-tools/iter-tools/blob/api-md/API.md

conartist6:
API.md needs some sort of cue to indicate that it is generated and the appropriate way to change it. I think a badge would make sense.

conartist6:
Additionally the DOCME files will be used to power features of the website. The website is coming soon, but at the moment `7.0.0-rc.1` is higher priority.

PR #290, opened 2019-09-16
Add Node.js workflow

KSXGitHub:
Resolve https://github.com/iter-tools/iter-tools/issues/273

### DIFF
--- a/.github/workflows/node.yml
+++ b/.github/workflows/node.yml
@@ -1,27 +1,67 @@
 name: Node CI
 on: [push, pull_request]
 jobs:
-  test:
+  test-js:
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
-      matrix:
-        node-version:
-          - 10.x
-        npm-script:
-          - test:js
-          - test:types
-          - lint:js
-          - lint:ts
     steps:
       - uses: actions/checkout@v1
-      - name: Use Node.js ${{ matrix.node-version }}
+      - name: Use Node.js 10.x
         uses: actions/setup-node@v1
         with:
-          node-version: ${{ matrix.node-version }}
+          node-version: 10.x
       - name: Install dependencies
         run: npm ci
       - name: Test
         env:
           CI: 'true'
-        run: npm run ${{ matrix.npm-script }}
+        run: npm run test:js
+  test-types:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+    steps:
+      - uses: actions/checkout@v1
+      - name: Use Node.js 10.x
+        uses: actions/setup-node@v1
+        with:
+          node-version: 10.x
+      - name: Install dependencies
+        run: npm ci
+      - name: Test
+        env:
+          CI: 'true'
+        run: npm run test:types
+  lint-js:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


KSXGitHub:
This is what I meant by shorter feedback loop:

checks screenshot

We will see every individual check at once.

conartist6:
Let me know when this is ready to go

KSXGitHub:
> Let me know when this is ready to go

Yes.

.github/workflows/node.yml
```diff
@@ -0,0 +1,29 @@
...
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        node-version:
+          - 10.x
+        npm-script:
+          - test:js
+          - test:types
+          - verify
```
conartist6: Don't run `verify`. Verify just runs all these other steps that you're calling individually, so there's no point in doing that, particularly if Travis is also running the complete verify script.

.github/workflows/node.yml
```diff
@@ -0,0 +1,29 @@
...
+on: [push, pull_request]
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        node-version:
+          - 10.x
+        npm-script:
```
conartist6: This appears to be an abuse of the actions API. Matrix is intended to allow you to choose node and OS versions. If you want to run different kinds of tests, you should be defining multiple jobs.

Why do you need to define multiple jobs? ALL the jobs which you have listed here require the code generator to be run first. Only the jobs API supports the setup you need, allowing you to set up a sequential step, after which several different kinds of tasks are executed in parallel. That API is documented [here](https://help.github.com/en/articles/workflow-syntax-for-github-actions#jobsjob_idneeds)

KSXGitHub: Very well, I will do as you wish when I'm back at my computer.

KSXGitHub: > ALL the jobs which you have listed here require the code generator to be run first. Only the jobs API supports the setup you need, allowing you to set up a sequential step, after which several different kinds of tasks are executed in parallel.

I just discovered that is not how it works ([I tried on a different repo](https://github.com/ksxnodemodules/ts-pipe-compose/runs/228072115)): Each job run on their own virtual machine, completely isolated from each other, this is evidence by the fact that `runs_on` is required for every job. That means that if a "code generator" job already generate the code, other jobs won't have access to the generated code even if it depends on "code generator".

KSXGitHub: So, do you still want to divide the matrix into multiple jobs?

KSXGitHub:
@conartist6 It is done.

We barely utilize GitHub Actions for faster testing, not its full potential: Such as creating a bot to automatically comment upon test completion, review a pull request, build documentation website and publish it, build package and publish it, etc.

As for https://github.com/iter-tools/iter-tools/issues/293#issuecomment-533199411, I can't because you haven't created it yet.

conartist6:
I have now switched to actions. I didn't do anything quite this complex. Currently I'm only testing on a single node version (v12), but I may yet decide to create additional jobs (or workflows) to test on older node versions.

PR #296, opened 2019-09-27
Change cursor to window and trailingWindow

conartist6:
I was improving the cursor documentation, when I realized that this function would more accurately be termed "window". I think this describes what it does quite well. A window suggests a frame of fixed size, which is what we have. The way I interpret cursor mirrors how cursors work in text. Usually a cursor could be before the first element or after the last one, and would only ever be between two elements at a time.

Also while I was refactoring, I realized that up until now, cursor has been completely backwards :'(. If you specify, `trailing: true` you get a leading cursor, and if you specify `trailing: false` you get a trailing cursor. The tests are all written backwards too.


conartist6:
In fact we already referred to the thing moving over the iterable as the window in documentation.

conartist6:
Ah, and I made it possible to specify the all-important size positionally.

PR #298, opened 2019-09-27
Split find into find and findOr

conartist6:


conartist6:
This will better match `firstOr` and `lastOr`.

PR #299, opened 2019-09-29
Forbid empty partial applications

### LINKED ISSUE #288: Variadics can't take no iterables
I'm struggling to figure out how to think about this problem.

The methods it affects are `collate`, `zip`, `zipAll`, `equals` (in next PR).

Currently these methods return a currying function if no iterables are passed to them. This makes sense, except when using a spread operator, e.g.
```js
zip(...iterables)
```
In this case if iterables is `[]`, you get the function back.

conartist6:
Fixes #288


conartist6:
To be perfectly clear, the thinking here is that this will usually be the result of a spread (e.g. zip(...arrs)`, where `!arrs.length`), and it will usually be a mistake which results in an error anyway. In that case this message is better than `Uncaught TypeError: function is not iterable`.

conartist6:
Need to add a bit to the CHANGELOG about this. Forgot that bit.

PR #300, opened 2019-09-29
Add arrayFrom and arrayFromAsync

conartist6:
`toArray` and `asyncToArray` are now aliases, but I have no intention of deprecating them. I still need them internally because they have `async` as a prefix (as opposed to a suffix), which means that `$toArray` can be used to reference the sync or async version.

I'm a little bit on the fence though about whether `arrayFromAsync` or `asyncFromArray` is a better name. I think from a purely linguistic perspective, `arrayFromAsync` is the most correct name to describe what the method does. That said, every other method has async as a prefix. Will coders remember to put async at the end? I think I'm OK with being a little bit inconsistent here because I plan to keep both options for the long run. If users value consistency of async being a prefix, there is no reason they should not continue to use `toArray` and `asyncToArray`.

Once I have created a repo of iter-tools lint rules, a rule could go into that repo that configurably prefers one name or another in order to help avoid the confusion that might be cause if the names were to be mixed in the same codebase.


conartist6:
https://github.com/iter-tools/iter-tools/blob/array-from/API.md is the thing to take a look at, specifically the "Consume an iterable" section.

PR #302, opened 2019-10-01
Use eslint to lint typescript and enforce prettier. Use lint-staged

conartist6:


conartist6:
This should significantly help reduce unnecessary work done on the developer's machine, while also reducing unnecessary cycles caused by accidentally pushing changes which are not formatted correctly with prettier.

conartist6:
Ah, I also had to start checking in the autogenerated type specs again, because they can contain lint errors which can't possibly be caught anywhere else, and if the linter only runs on changed files it needs the files to be tracked in git to know if they've changed.

PR #311, opened 2019-10-12
Add lint rule to forbid focused tests.

### LINKED ISSUE #309: Add a lint rule forbidding `it.only` and `describe.only` in tests

sowmiyamuthuraman:
Fixes #309

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,5 +1,6 @@
 module.exports = {
-  extends: ['standard', 'plugin:prettier/recommended', 'typescript'],
+  extends: ['standard', 'plugin:prettier/recommended', 'typescript', 'plugin:jest/recommended'],
+  plugins: ['jest'],
   parser: 'babel-eslint',
   parserOptions: {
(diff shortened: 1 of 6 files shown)


conartist6:
Could you try including 
```
{
  "extends": ["plugin:jest/recommended"]
}
```
in `.eslintrc` and see how close we are to meeting those rules? If there are rules we clearly don't meet then you could turn them off one by one in the `rules` config, but for the rules we do meet already I'd rather have them on.

sowmiyamuthuraman:
> Could you try including
> 
> ```
> {
>   "extends": ["plugin:jest/recommended"]
> }
> ```
> 
> in `.eslintrc` and see how close we are to meeting those rules? If there are rules we clearly don't meet then you could turn them off one by one in the `rules` config, but for the rules we do meet already I'd rather have them on.

@conartist6 i have made the suggested changes and amended to the same commit

sowmiyamuthuraman:
@conartist6 can you please review the changes?

PR #312, opened 2019-10-14
Codecov integration 

JacobMGEvans:
- reverted branch for travis file
- started codecov setup
- also deleted coverage badges and the old scripts
closes #312


JacobMGEvans:
I added configurations and codecov to the packages in accordance with the documentation for codecov environments here: https://github.com/codecov/example-node#jest

JacobMGEvans:
Very exciting. Once I found some examples and focused on this project I was able to figure some things out... Thank you for the patience and understanding so far too BTW

JacobMGEvans:
@conartist6 have you had a chance to review the PR?

package.json
```diff
@@ -27,7 +26,7 @@
     "build:es2018-cjs": "cross-env BABEL_ENV=cjs babel --config-file ./babel-es2018.config.js -x .mjs src -d es2018",
     "build:copyts": "copyfiles -f src/*.ts es5 && copyfiles -f src/*.ts es2015 && copyfiles -f src/*.ts es2018",
     "create:method": "node ./generate-yo method",
-    "verify": "run-s generate verify:* build test",
+    "verify": "run-s generate verify:* build coverage test:js",
```
conartist6: This doesn't run the typescript tests anymore. Change `test:js` to `test:ts`

package.json
```diff
@@ -8,12 +8,11 @@
   "typings": "es5/index.d.ts",
   "sideEffects": false,
   "scripts": {
-    "test:js": "jest",
+    "test:js": "jest && codecov",
```
conartist6: `test:js` shouldn't run coverage. It slows down test running and makes stack traces unintelligible, which is not what I want for the script that people will be running to validate their changes during local development. Also during local dev there's not much point in uploading the results using the `codecov` command. In fact I'd say only the `verify` script aught to do that, since that's the script I expect CI to run.

package.json
```diff
@@ -132,5 +132,9 @@
   },
   "resolutions": {
     "babel-core": "7.0.0-bridge.0"
+  },
+  "jest": {
+    "coverageDirectory": "./coverage/",
+    "collectCoverage": true
```
conartist6: Remove this line. You want to instead invoke `jest --coverage` as your `coverage` script.

JacobMGEvans:
If you would that would be great! Thanks for the opportunity and all the
help along the way. 🤩

JacobMGEvans:
> It looks pretty good. I have some last feedback, but this is just stylistic stuff. If you want I can amend these changes onto your commit for you.

If you would amend those things that would be awesome. @conartist6 
Thanks again!

PR #314, opened 2019-10-15
Update package build

conartist6:
Fix some issues with the package build that had cropped up since I implemented the generator system.


conartist6:
@KSXGitHub is this good enough for you in terms of types? I tried importing the package in a test typescript project and it the types seemed to work fine.

conartist6:
If you want to try it out, you can make a test project and install the build output from this PR, which I've just uploaded:
```
npm i https://github.com/iter-tools/iter-tools/files/3730188/iter-tools-6.2.3.tar.gz
```

KSXGitHub:
I see `T = any`, `extends any`, and `extends any[] = any[]` in multiple places, I do not think it is necessary. I guess you did it because you want `any` to be used by default instead of `unknown`, but I would prefer `unknown` as it is harder to make mistakes with.

conartist6:
I decided to provide default values because the default if nothing matches is (was?) `{}`, which is confusing to me on its own but tends to lead to even more confusing error messages. Maybe newer versions of typescript have changed that behavior though?

I do see the upside of `unknown`, though. Just a curried `map` call makes it easy to see.
```
// with T = any
map(x => ++x)(["a", "b", "c"]) // no error
// with T = unknown
map(x => ++x)(["a", "b", "c"]) // error on ++x
map((x: string) => ++x)(["a", "b", "c"]) // error on ++x
map((x: number) => ++x)(["a", "b", "c"]) // error on iterable
```

The one thing that I know will be a bit tricky is the tests, which generate the `auto.spec.ts` files to check the TS types, but will also be used to test Flow types in the future. I think I'll need a babel plugin to strip type definitions off of arguments since we'll be wanting to test that flow works without them, but that should be doable.

KSXGitHub:
> I decided to provide default values because the default if nothing matches is (was?) `{}`

It's `unknown` now.

conartist6:
So do you think it makes sense to be explicit about making the default `unknown` in order to support older versions of TS, or should we just remove them and say that we intend to support TS v3?

KSXGitHub:
`unknown` doesn't exist on older TS so removing `any` is the only way

conartist6:
Ah right. OK, I'll do that then.

KSXGitHub:
Yes

PR #315, opened 2019-10-19
Remove = any defaults from type params

conartist6:


conartist6:
Looks like the only `extends any` is in the `size` implementation, where it is appropriate.

There's also this, which you wrote:
```ts
type ProductReturnElement<Args extends Array<Iterable<any>>, Holder extends any[] = []> = {
  empty: Holder;
  many: ((...a: Reverse<Args>) => any) extends ((a: infer Last, ...b: infer ReversedRest) => any)
    ? ProductReturnElement<
        Reverse<ReversedRest>,
        Prepend<Holder, Last extends Iterable<infer T> ? T : never>
      >
    : never;
  infinite: Args extends Array<Array<infer T>> ? T[] : never;
}[Args extends [] ? 'empty' : IsFinite<Args, 'many', 'infinite'>];
```

Do you want that to change? I can't say I understand everything going on there.

We als have a quite a few more instances of `any` in the types that are not associated with generics, e.g. the `shouldFlat` callback and methods that search for values, such as:
```
declare function startsWith(value: any, iterable: SourceIterable<any>): boolean;
```

I don't think those are necessarily a problem.

conartist6:
Obviously we could try to save people from looking for things we don't think they'll find in an iterable, but we aren't generating any more information about she shape of the data at any point and there's no runtime harm in looking for something you won't find.

KSXGitHub:
> ```ts
> type ProductReturnElement<Args extends Array<Iterable<any>>, Holder extends any[] = []> = {
>   empty: Holder;
>   many: ((...a: Reverse<Args>) => any) extends ((a: infer Last, ...b: infer ReversedRest) => any)
>     ? ProductReturnElement<
>         Reverse<ReversedRest>,
>         Prepend<Holder, Last extends Iterable<infer T> ? T : never>
>       >
>     : never;
>   infinite: Args extends Array<Array<infer T>> ? T[] : never;
> }[Args extends [] ? 'empty' : IsFinite<Args, 'many', 'infinite'>];
> ```

I wrote this, because I think it would be useful. But if we were to generate documentation website from TypeScript definition, we have to change this to a simpler definition sooner or later.

conartist6:
Yeah I do think it helps when people can understand the type definition. I'm not certain that I intend for the documentation website to list the typescript types at this point. I rather like my documentation I've created, though the website could certainly make it much prettier and easier to navigate.

One thing about that `ProductReturnElement` type though is that I think it claims the output at each step is an array, which is not the case. This of course is more or less necessary, since typescript has no concept of an iterable tuple. Do you think there's any chance of convincing them to create such a concept for us? I can't imagine that they won't eventually need to do it.

KSXGitHub:
> Do you think there's any chance of convincing them to create such a concept for us? I can't imagine that they won't eventually need to do it.

The thing about this "iterable tuple" is, it isn't sound: Let's say we have a `(0, 1, 2, 3)` tuple, after iteration (consume), it must turn into `()`, not to mention possible shared references which are unaware of the consumption. I can't imagine them do this, ever.

conartist6:
It isn't impossible to understand generators in TS. You know the second iteration will be empty if you have `[Symbol.iterator](): this;`. I don't think TS needs to understand every way  of consuming an iterator, such as `next()` calls, but it should specifically support array spreading and iterable destructuring, both of which should be safe from shared references since they have control until they're done consuming the iterable.

KSXGitHub:
How are you going to deal with this code:

```typescript
const a = range(3) // should be (0, 1, 2, 3)
const b = a // should be (0, 1, 2, 3)
const c = [...a] // should be [0, 1, 2, 3]
type B = typeof b // should be (), somehow
const d = [...a] // should be []
const e = [...b] // should be [], somehow
```

conartist6:
Still don't think it's a complete deal-breaker. You don't need the type system to understand this any better than it currently does, you just need to it to understand the difference between this and what it can understand.

PR #329, opened 2019-11-15
Keys values entries

conartist6:
Our iterables now have three additional iterators, `keys`, `values`, and `entries`. None of the default iterators are changed. By default keys are numeric indexes, with the exception being `group` and `groupBy` where the keys are the grouping keys.

That reminds me now we can make a `keyBy` method.


conartist6:
Ultimately I envision an API where callbacks that are passed arguments `(value, i)` would instead be given arguments of the form `(value, key)`, where `entries` would be used to get `value` and `key`. This is essentially what Immutable.js does. I think the correct way to offer this would be through an additional package though (Sequin, I'm thinking) because if iter-tools has that kind of high level behavior it becomes less suitable for low level usage.

conartist6:
I'm still unsure of these changes. This may be the rare case where I reject my own PR. If my plan is for usage of this functionality to be through a sugaring layer, sequins, then why include the functionality here at all?

chuckwondo:
@conartist6 I presume you've chosen to implement `groupBy` to return an iterable to keep in line with what I gather to be the general idea of lazy evaluation for this library. This is one of the reasons that I'm am happy to have found this library, so thank you for this library!

However, it would seem to me that in this case it would make more sense to not stick with laziness because the result is not only counter to what is commonly found elsewhere (in fact, _everywhere_ else that I've personally encountered it), but perhaps even not particularly useful (but I might well be wrong here).

Specifically, it doesn't truly group _everything_ in the iterable by key, but rather groups only running sub-sequences by matching key. I don't know how useful that is, but if it is useful, consider using a different name, as I imagine that most people would expect `groupBy` to return a mapping from key to _all_ values in the iterable with the same key, not just the values in the current sub-sequence.

In summary, I'd like to suggest that you consider making `groupBy` (whether or not you pull it into a `sequins` package -- nice name, by the way) return what is arguably what most would expect (a mapping containing distinct keys mapped to _all_ corresponding values for each key), and renaming this current implementation (if it is indeed useful to keep) to something else, such as `groupSpansBy` or `groupRunsBy` or something that indicates that grouping is occurring in "spans" or "runs" of matching keys.

chuckwondo:
@conartist6 Sorry, I just looked at Python's `groupby` in `itertools` and see that you are mimicking the behavior there. I also see that the description there indicates that a solution to the "problem" I described in my previous comment would be to sort the input before invoking `groupby`. Given that this library is inspired by Python's `itertools` library, then I suppose it makes sense to stick with your implementation, although I must say I'm a bit torn on this behavior, as `itertools` is the only place that I've noticed such behavior. Are there other languages/libs that take this approach to `groupBy`?

conartist6:
It's unique to iterators. You want to support this version of `groupBy` because it is what is possible to do lazily. Almost everyone will probably also at some point need to do groups on an unsorted iterable though. In that case it is always possible to use [lodash.groupBy](https://www.npmjs.com/package/lodash.groupby), but it's a [12 kilobyte](https://bundlephobia.com/result?p=lodash.groupby@4.6.0) package, which is just nuts.

I think I'd be open in this case to adding the sugar to this library to convert to an array and sort it before applying `groupBy`. What would such a function be called? `groupAllBy`? `groupArrayBy`? `fullGroupBy`? `greedyGroupBy`?

I'll think a little more on it. `sort` also is only well defined over arrays, but maybe not a bad idea to add the sugar to it anyway, just so that I don't have to tell people that they need to have iter-tools and lodash.

chuckwondo:
I would suggest not performing sorting, but rather doing the iteration internally since there's no need to bump the algorithm from O(n) to O(n log n) to achieve the same result as a mapping.

It's a shame the lazy function is named `groupby` in `itertools` since elsewhere "group by" is typically not lazy (and "keys" are unique). It would have been nice if a slightly different name were used for the lazy version since it produces different results than the typical "eager" version (except when the iterable produces values in order).

Among the names you suggested, perhaps `groupAllBy` is a good name for the "usual" implementation that returns a mapping instead of an iterable. What about `groupByUnique`?

Speaking of "unique", not to derail this thread, but have you considered adding a "unique" or "distinct" function? If you care to discuss, I'll open a separate issue for discussion.

conartist6:
Good points. `group` exists and has essentially that function, but of course it's localized as `groupBy` is.

conartist6:
@chuckwondo I happened back on this discussion and thought you were right about it being a bad idea to take the same name for something with a different implementation than users expect. I chose the name `splitGroupsBy` instead. I have to think some more over how and where it might be appropriate to add an implementation of the standard `groupBy` method.

PR #330, opened 2019-11-17
splitAt accepts negative indexes. spliterators

conartist6:


conartist6:
I don't really like the number of caveats that `asyncSplitAt` needs. I'm considering turning it into `asyncSplitAt` and `splitAsyncAt`, where `asyncSplitAt` returns an async iterable of two async parts, and `splitAsyncAt` returns a sync iterable of two async parts.

This makes it easier for docs to tell people what to do, but will it be hard for people to remember? And is there really any use case for `asyncSplitAt`?

conartist6:
Oops, looks like the `DoneProxy` isn't actually being used...

Am I missing a better solution to this problem?

conartist6:
I've also duplicated the functionality of caching items between WeakExchange and PartManager : (

conartist6:
OK, a vastly reworked version of this PR will be up soon which has much better code reuse yet also avoids dependencies on complex data structures and methods where they aren't really needed. I've also decided that I'm shooting myself in the foot (and allowing others to do so as well) by caching parts when they are taken/consumed out of order. Instead I'll now just throw an error if the user tries to take items from the first part after they have requested the second part. This behavior will be for all methods which split, including the `group` family.

One of the reasons I'm comfortable doing this is that if there is a legitimate need to consume parts in arbitrary order, it's quite simple for a user to cache the parts themselves by writing something like `map(toArray, splitOn(null, iterable))`. This has the added benefit of making resource usage explicit both when reading and writing the code, and ensures that when you don't see explicit resource usage there isn't any.

conartist6:
OK the new changes are up. Not fully done yet -- I still need to update the CHANGELOG and some method docs, as well as handle return conditions for `splitAt`.

conartist6:
I think this is ready. It now is a significant increase in coverage as opposed to a significant decrease.

Issue #372, opened 2020-10-29
batch: batches should be iterables?


conartist6:
Currently the batches returned by the `batch` method are arrays. This is undocumented, and in general I think we are trying to avoid ever exposing arrays as opposed to iterables. Supposing the user was creating very large batches (e.g. 1k, 10k) the perf characteristics would not be as expected.

To note though, there are some places where we do yield arrays and probably always will. `zip` is an example. Arrays seem non-negotiable for `zip` results because they are the only way we'd be able to write type definitions that capture what's really going on -- which I notice we haven't done yet. E.g. `zip([1], ['a'])` should result in `Iterable<[number, string]>`. We already do this for `interleave`, and I believe it is going to require less repetition to code using typescript v4's variadic tuple types once we are able to use them.

PR #374, opened 2020-11-08
Spliterate

### LINKED ISSUE #365: Peekerator
I'm trying to build a streaming parser, and I find that I need something very similar to the `InterleaveBuffer` that already exists. As long as the buffer was also capable of storing more than one item this would be used to implement lookahead.

The only problem I foresee is that I'm trying to build a streaming parser (for leading `/**/` comments) for use in macrome, which is soon going to be used to build `iter-tools@7`, meaning iter-tools would always need an older version itself to build.

conartist6:
Fixes #365, #372.


conartist6:
Travis is doing something weird. It says it has received the builds, but it hasn't moved them to the queue. I'm just going to run the verify script locally. It passed.

PR #376, opened 2020-11-11
Munge dollar imports at the Program level

conartist6:


conartist6:
Ran verify locally again as Travis is behind again.

PR #381, opened 2020-11-16
remove es2015, add explode.macro

### LINKED ISSUE #343: Stop shipping corejs (and codecov)

conartist6:
Fixes #343.


conartist6:
Next question, can I find a way to stop my implementations being completely copied between cjs and non-cjs files, and is it worth whatever cost?

conartist6:
Fixed a problem I created earlier when I deleted the dependency on `@babel/runtime`. The tests had continued to pass because some other module had been depending on `@babel/runtime` in a way that it was hoisted to the root of `node_modules`. I should figure out how to avoid that kind of false positive in the future.

conartist6:
Verified locally. I'm going to merge this as it fixes a breaking bug and is a good stepping stone to further changes I am considering.

PR #382, opened 2020-11-20
Separate es and es5 packages

conartist6:
Splits the two folder container different transpilation levels (`/` and `/es`) into separate packages (`@iter-tools/es` being the new one). This is described in more detail in the changed `README.md`. Each package supports commonjs and es imports, but instead of having implementation code copied between the commonjs and es modules some clever (and very custom) transpilation ensures that the bulk of the code is not duplicated. Oh, and for bonus points the repo root itself should now be a valid es package : ).

I have to test it all thoroughly and make sure it works like I think it does. Right now we don't have any tests pointed at the `dist` folder, but I think it would be wise to ensure that our test suite can execute on it.


conartist6:
For anyone keeping count, that means an approximately 6x reduction in the amount of code we ship.

conartist6:
Oh and I still have to put my `/*#__PURE__*/` annotations back. Another win: I think even the commonjs code will support tree-shaking. This should by and large not be of consequence since most bundlers are correctly configured to prefer module code over non-module code, but you never know.

conartist6:
I've decided to put the pure annotations in the sources since this is the only way it will be possible to tree shake when using the repo as a package, and I want using the repo as a package to have first class support. I have therefore written my first eslint rule to ensure that this is done consistently. Thanks to kaicataldo on discord for the suggestions as to how to write rules without having to publish them as packages!

conartist6:
Pure annotations done. Jest is now using es modules natively, and it is possible to import the repo and from a node es-module and it just works. I need to build some infra to test the packages that I generate. I may want to delay that to a separate PR though as I think I plan to shift to github actions. Travis has been taking hours to run our actions, which is more than I am willing to wait.

I'll have to think about the dimensions I want to test on now. I think the ideal thing would be for local dev to have `jest` transpile on the fly while ci verification is done against the built packages. I'll just make sure it's easy to run either script locally for debugging, e.g. `npm run test` and `npm run test:built`.

conartist6:
I think I'm done. I just need to test it all and make sure it works in all the environments that I intend to support. Jest is now capable of testing the build code (in `/dist/es` and `/dist/es5`) and its normal tests now consume `/src` through the root of the repo (i.e. by reading `package.json`), so the tests pretty much do exactly what a real consumer would. Still, it's worth checking.

conartist6:
Verified locally. Going to merge this so I can get on to unbreaking the build by moving to github actions.

PR #384, opened 2020-11-29
Move from Travis to Github Actions

conartist6:


conartist6:
I tried yarn 2, but I did not like the amount of ugly it injected into git. I can see how the gains might be worth it for a package with a complicated dependency structure, but this is a zero-dep package that just needs to be built.

conartist6:
For the record I ditched npm because it was responsible for breaking the build. It has all kind of problems with git dependencies. I filed [this issue](https://github.com/npm/cli/issues/2256), and there's a slew of others already open. Yarn doesn't seem to have that issue.

PR #389, opened 2020-12-01
Documentation improvements

conartist6:
Changed `item` to `value` everywhere, as this is the correct terminology. `item` had also been used to describe the `{ done, value }` tuple. The current term for this is `step`.


conartist6:
I also pulled the cookbook as it was mostly either outdated or described patterns we would not want to encourage. I'm gathering new ideas for real problems that we can show solutions to in the cookbook.

PR #392, opened 2020-12-02
Fix missing pure call sites by rewriting lint rule

conartist6:
The new rule does a much better job. It can even figure out some things we don't do yet, and it found problems I hadn't even thought of in addition to some that I had. That said, it sure isn't perfect. I don't even understand all the conditions I decided just to skip over.


conartist6:
Also this is super important to get right in linting because the failure mode (inability to drop a file during tree shaking) is pretty difficult to test (and is currently not tested).

PR #394, opened 2020-12-05
Add __methods to bypass currying

conartist6:


conartist6:
Also fixes a bug in `explode.macro` uncovered by adding better tests, but unfortunately those tests had to disabled because they rely on self-reference support in resolve which does not exist yet. See: https://github.com/browserify/resolve/pull/224. I could probably mock the necessary code, but I don't expect explode.macro to change often from now on, there's a workaround for manually running the test, and it looks like the work on resolve is close to being done.

Issue #224, opened 2019-05-04
Discussion: how many async methods should offer parallelism?


conartist6:
Currently map and filter offer parallelism, however there are several other functions for which that optimization would be relevant. It is relevant when there are serial awaited calls to async functions mixed in with consuming an async iterable. Other [async] methods that meet this criteria are:
* some
* every
* takeWhile
* dropWhile
* find
* groupBy
* partition
* multiPartition
* reduce
* tap

There are several things to consider before we forge ahead here. How hard will it be to adapt our ParallelMapper code to adapt to early termination? Is it OK to overconsume the source? (For example a parallel find might get the first 5 items only to discover that the first item matches the predicate.)  Are there perf reasons not to make code use the ParallelMapper if it has concurrency 1? Certainly page weight would be one such concern, the only solution to which would be to create new and separate functions like `asyncMapParallel`. As a note that would not be a breaking change, since we have not yet released any functions with parallel implementations.

sithmel:
We have a function that consume the iterable eagerly. And it is asyncBuffer.

The intention with map, filter and flatMap, was allowing to run their function argument in parallel.

For example:
```js
const superagent = require('superagent')

asyncMap((n) => {
  superagent
  .get(`/items/${n}`)
}, range(10))
```

conartist6:
Right. Hence the list of functions that can run an async function argument, and would potentially benefit from doing it in parallel.

sithmel:
@conartist6 yes.

Although, now that I see that with fresh eyes, I wonder if the numeric parameter was a mistake.

Perhaps we should have just implemented asyncMap like a chain of promises, like the ParallelMapper simplified. And use asyncThrottle to limit the parallelism.

conartist6:
After everything we put into it the first time, I'm relatively certain there's no simpler way to rewrite asyncMap. And anyway a chain of promises is exactly what ParallelMapper is!

sithmel:
Indeed, I meant ParallelMapper without number.

But your point stand still. I think all these methods can benefit of parallelization

sithmel:
Hang on, ignore what I wrote about "ParallelMapper without number". I am not sure it makes sense :man_facepalming: 

Let me double check

sithmel:
Yes I confirm. What I said doesn't make sense.
So for me it is definitely a *yes*.

In another news, running the benchmarks, ParallelMapper is faster than it should be. I suspect the concurrency is off by one. I am investigating

[...]

PR #401, opened 2020-12-06
Allow parallel runner to work with sync iterables

conartist6:
Most likely that would mean sync iterables of promises, but now it won't crash even if the sync iterable yields something that is not a promise.


conartist6:
Wow OK, this one changed a lot since I started working on it. It now addresses the issues discussed in #224, and allowed me to delete a large amount of untested code that was wrong and replace it with a small amount of properly tested code that is right. 🎉

PR #407, opened 2020-12-10
IterableIterator types: improve and export

conartist6:
Helps avoid the madness that is `iter[Symbol.iterator]()[Symbol.iterator]()[Symbol.iterator]()`. Of course nobody should be writing the syntax out that explicitly, but I still think that if it happens it's a mistake, so typescript is going to point it out now.

Also by upgrading to `typescript@3.8` we got the `export type` syntax, so now users will be allowed to declare variables which will hold values generated by iter-tools without any loss of fidelity.


conartist6:
Ah yes, and the names changed. I think the names in the code finally reflect the language I use everywhere.

PR #409, opened 2020-12-10
Remove combinatorics

conartist6:
These are going into their own package. I beg forgiveness for telling users to use a package that doesn't exist yet. I've got a bunch of it written, I just need to make a few more tweaks so that `iterableCurry` is a public API.


conartist6:
Oof, looks like I messed up and left out some of the consume changes from my last PR.

conartist6:
Oh god, no, I just accidentally deleted consume entirely. Thankfully I decided not to overlook the inconsistency.

PR #415, opened 2020-12-12
windowAhead

conartist6:
I decided on this change because once again I could never remember which method was which. `leadingWindow` always made me wonder "is the window on the leading edge of the iterable, or is the current value on the leading edge of the window?".

Also it works better alphabetically.


conartist6:
I changed the argument order because it makes way more sense to partially apply the options then choose your size. I called this a removal in the change log just because it's hard to make sure people will see changes to a method that's been both renamed and changed. Which name do you attribute the changes to?

PR #429, opened 2021-03-21
Add typings for zip with multiple source types

canac:
If there's a way to do this generically in TypeScript, that would be awesome! But until then, this PR adds typings to `zip` for the two and three arguments cases so that the sources can be of different types.


src/impls/$zip/__tests__/$zip.test.ts
```diff
@@ -26,4 +26,27 @@ describe($`zip`, () => {
...
+        [3, 'c'],
+      ]);
+
+      const threeSources = $zip(
+        $wrap([1, 2, 3]),
+        $wrap(['a', 'b', 'c']),
+        $wrap([true, false, true]),
+      );
+      expect($await($toArray(threeSources))).toEqual([
+        [1, 'a', true],
```
conartist6: I think if you do `[1, 'a', true] as const` typescript will understand this as a tuple type which will ensure that this test won't pass  unless the types you added are present and working.

conartist6 (commented):
Would you copy these changes to the zipAll method as well? Otherwise it looks good.

canac:
@conartist6 I went ahead and made typings spec tests. To me it makes the most sense to keep those new spec tests and remove the other tests that I added because the goal is to test the typings. The functionality is already tested, so we don't need to duplicate those tests with the special cases for different argument types. But let me know how to proceed.

src/impls/$zip/__tests__/$zip.test.ts
canac: I'm not familiar with using `as const` in this context, but I went ahead and added it for you.

conartist6:
I agree with you -- just keep the `spec.ts` tests you wrote. If you get rid of the `.test.js` changes I'll approve.

Also for the record I don't think you're using `as const` quite right. The docs on it are [here](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html#const-assertions). In particular I think each tuple in the array needs an `as const`. It doesn't do anything to the outer array the way you used it because all the elements are already of type `Array<T1|T2|T3>`.

canac:
Removed the functionality tests as requested.

conartist6:
Thanks for your contributions! We didn't end up with any runtime changes so I think this can be `v7.1.1`.

PR #430, opened 2021-03-23
Unwrap promises in AsyncIterableIterator

canac:
Here's an initial attempt at making sure that we get types like `AsyncIterableIterator<number>` instead of `AsyncIterableIterator<Promise<number>>`.


conartist6 (approved):
This is great!

PR #434, opened 2021-05-06
Fix `collate`'s use of the `compare` function

### LINKED ISSUE #433: collate/asyncCollate incorrectly compare elements for ordering
`collate` uses the `compare` function in the wrong "direction" for ordering (merging) elements. In `src/impls/$collate/[async-]collate.js`, the `[async]byComparison` function compares this way:

```javascript
if (!peekr.done && compare(candidate.value, peekr.value) < 0) {
  candidate = peekr;
}
```

That ends up assigning `candidate` the `peekr` with the _larger_ value rather than the one with the _smaller_ value.

I believe one possible fix is the following:

```javascript
if (!peekr.done && compare(peekr.value, candidate.value) < 0) {
  candidate = peekr;
}
```

The error is evidenced by this example, taken straight from `API.md`:

```javascript
collate((a, b) => b - a, [6, 5, 2, 1], [4, 3]); // Iterable[ 4, 3, 6, 5, 2, 1 ]
```

Notice that `b - a` is what _should_ work for _descending_ ordering, but it does not.  However, by changing the comparator to use `a - b` (intended for _ascending_ ordering), the correct _descending_ ordering is produced.

Further, the related tests pass for producing _ascending_ results because they use `b - a`, but should use `a - b` instead. In other words, for _ascending_ order, the comparator function must return a value less than 0 when `a < b`, which in the case of the tests means they should use `a - b`, not `b - a`.

chuckwondo:
Fixes #433

### DIFF
--- a/API.md
+++ b/API.md
@@ -1195,7 +1195,7 @@ See [splitWhen](#splitwhen)
 Combines values from each `source` in `sources` into a single iterable, peserving the ordering of values within each `source`. Collate uses `comparator` to establish a partial ordering of values at the head of each `source`. At each step it yields the lowest value in the ordering then recomputes the ordering.
 
 ```js
-collate([1, 2, 5, 6], [3, 4]); // Iterable[1, 2, 3, 4, 5, 6]
+collate((a, b) => a - b, [1, 2, 5, 6], [3, 4]); // Iterable[1, 2, 3, 4, 5, 6]
 collate((a, b) => b - a, [6, 5, 2, 1], [4, 3]); // Iterable[6, 5, 4, 3, 2, 1]
 ```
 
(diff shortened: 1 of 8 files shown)


conartist6:
You need to modify `src/impls/$collate/README.md` as well. Otherwise thanks for putting this together!

I *think* I'm OK with putting this out in semver-patch. I don't know of anyone using this API, and I think if anyone else had tried to use it they would have noticed the problem.

chuckwondo:
Fixed `src/impls/$collate/README.md`. Let me know if that's the update you meant.

conartist6:
It is. You're going to have to rebuild again so that the change gets rolled into the root level `API.md` file (it'll fail the build in a moment). Also could you squash your commits together? With those fixes I'll merge.

chuckwondo:
Ah, sorry. Done.

conartist6:
Released as `7.1.3`

PR #449, opened 2021-10-25
Fixed typo in pipe and compose docs

Pyrolistical:
The docs for `pipe` used the expression `c(b(a))`, which means `a` as a function is passed into `b`, which is misleading. What `pipe` really does is the result of `a(x)` is passed into `b`.

I've fixed the `pipe` and `compose` docs to make this more clear by showing `pipe` and `compose` return functions.

### DIFF
--- a/API.md
+++ b/API.md
@@ -2602,7 +2602,7 @@ Yields the elements from `source` in reverse order. `source` must be an array, s
 
 **compose(...fns)**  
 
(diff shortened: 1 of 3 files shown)


conartist6:
Thanks! That's a great improvement.

conartist6:
Sorry it's a bit odd. `API.md` is autogenerated. You just want one commit which changes `src/impls/pipe/README.md` and `src/impls/compose/README.md`, then run `yarn generate` and your changes will show up in `API.md`.

conartist6:
If the checks pass it's fine though, I'll just squash merge.

conartist6:
Oh sorry I didn't notice that it wanted me to give you permission for every workflow run. That seems a bit absurd. I've made that setting more permissive, so I think it should just run itself now.

PR #460, opened 2022-02-07
fix(take): finish the iterable as soon as requested items are took

GerkinDev:
`asyncTake` waits for the next iteration item before completing, which may never be yield. The break condition have been slightly modified & moved right after the yield to allow early exits.

While this issue should not happen for sync iterables, those modifications have been reflected on the sync version too.

### DIFF
--- a/src/impls/$take/__tests__/$take.test.ts
+++ b/src/impls/$take/__tests__/$take.test.ts
@@ -1,4 +1,4 @@
-import { $, $async, $await } from '../../../../generate/async.macro.cjs';
+import { $, $async, $await, $isAsync } from '../../../../generate/async.macro.cjs';
 
 import { $take } from 'iter-tools-es';
 import { $wrap, $unwrap } from '../../../test/$helpers.js';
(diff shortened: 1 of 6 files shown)


conartist6:
Looks good, but you created a bug with the `n === 0` corner case. Now we need  a bail-out check for `n===0` before the for loop. Also once you add that line you'll  need to test it for coverage purposes, though that should be as simple as adding a second line to that first test.

GerkinDev:
All set ! I used `<` & `>=` comparison operators to mitigate cases where non-int `n` would cause the iterable to never complete.

conartist6:
Oh, sorry, but I'd prefer to leave the comparisons as `===`. I have a separate system for validating argument types.

GerkinDev:
Oh yeah, I see this now, `$iterableCurry`. So, fixed. Nice tooling pipeline btw.

src/impls/$take/__tests__/$take.test.ts
```diff
@@ -10,4 +10,30 @@ describe($`take`, () => {
...
+    }),
+  );
+  it(
+    'completes immediately after taking the first n values',
+    $async(() => {
+      $async;
+      function* twoItemsThenNever() {
+        yield 1;
+        yield 2;
+        // Never yields 3.
```
conartist6: One last change. Could we just throw an error here instead of causing the test runner to time out?

GerkinDev: Yes, we could, but that case should never happens anyway as long as the implementation is correct. But you're the boss, on it

conartist6:
Thanks! I'm working on rebuilding the tooling around the `$` stuff right now actually, as some important parts of it (like watch mode) are broken. I do very much like that it allows contributors to edit once and end up with equally readable sync and async code though.

GerkinDev:
I know you don't asked me for my opinion but I think the philosophy is good.

The 2 small problems I see though is that
* the `async.macro.cjs` file does not really exports the symbols used in tests, and it gives annoying red underline in vscode
* and maybe the contributing documentation could be improved a bit. It took me a couple of minutes to understand where I should put my modifications and how the tooling interacts with the code.

conartist6:
I can fix both those things!

conartist6:
The fix is released in `v7.2.1`. I appreciate your help and patience!

PR #461, opened 2022-02-07
Minor improvements for contributors

conartist6:
@GerkinDev does this look good? I can't make all the vscode squigglies go away, but this gets rid of the ones you mentioned.


GerkinDev:
Didn't checkout, but LGTM, thank you ;) precious minutes saved here !

PR #480, opened 2022-06-10
add distinct operator

hillin:


API.md
```diff
@@ -437,6 +438,27 @@ append(4, [1, 2, 3]); // Iterable[1, 2, 3, 4]
...
+
+**distinct(selector, [source](#wrappable))**  
+**distinct([source](#wrappable))**  
+**__distinct([source](#iterable), ?selector)**  
+
+Yields only the first occurrance of each value from `source`. An optional `selector` can be used to specify a transform function which returns a key to distinguish the values.
+
+```js
+distinct([3, 1, 4, 1, 5, 9, 2, 6, 5, 3]); // Iterable[3, 1, 4, 5, 9, 2, 6]
+distinct((x) => x % 7, [14, 21, 9, 777]); // Iterable[0, 2]
```
conartist6: You'll want to make these examples smaller and simpler. Not everyone that reads these docs will even know english well, so it's pretty critical that it's easy to look at the example and understand it. I don't want people to have to do modulo 7 math in their head in order to understand how the method works.

Second, I think the example shows something incorrect about the implementation. I would expect:
```js
distinct((x) => x % 7, [14, 21, 9, 777]); // Iterable[14, 9]
```

src/impls/$distinct/$distinct.js
```diff
@@ -0,0 +1,23 @@
...
+    }
+
+    set.add(key);
+    yield value;
+  }
+}
+
+export const $distinct = $iterableCurry($__distinct, {
+  minArgs: 0,
+  maxArgs: 2,
```
conartist6: These numbers refer to the number of configuration arguments, i.e. those that are not iterables. You only have 1 arguments that is not an iterable, so `maxArgs: 1`

src/impls/$distinct/$distinct.js
```diff
@@ -0,0 +1,23 @@
+import { $async, $await } from '../../../generate/async.macro.cjs';
+
+import { $iterableCurry } from '../../internal/$iterable.js';
+
+$async;
+export function* $__distinct(source, selector) {
```
conartist6, line 6: The implementation looks good!

src/impls/$distinct/DOCME.json
```diff
@@ -0,0 +1,4 @@
+{
+  "type": "transform",
+  "signatures": [[{ "name": "selector" }], []]
```
conartist6: You don't need to define `signatures` -- the documentation engine will infer it in all but the weirdest cases. I'm guessing it just did something weird because `maxArgs` was off.

src/impls/$distinct/__tests__/$distinct.test.ts
```diff
@@ -0,0 +1,68 @@
...
+      }),
+    );
+  });
+
+  describe('when source has values and a selector is specified', () => {
+    it(
+      'only yield distinct values based on the selector',
+      $async(() => {
+        const source = [
+          {
```
conartist6: Let's make these objects one-liners so that more fits on the screen.

src/impls/$distinct/__tests__/$distinct.test.ts
```diff
@@ -0,0 +1,68 @@
...
+      }),
+    );
+  });
+
+  describe('when source has values', () => {
+    it(
+      'only yield distinct values',
+      $async(() => {
+        const source = [3, 1, 4, 1, 5, 9, 2, 6, 5, 3, 5, 8];
+        const result = [3, 1, 4, 5, 9, 2, 6, 8];
```
conartist6: This is OK as tests go, but I'd prefer that you use the smallest inputs that demonstrate particular behavior. I might do something like:
```
When input is empty: yields no values
null -> iter[]

When input contains only distinct values: yields values
iter[1] => iter[1]

When the input contains identical values: emits only the first value
iter[1, 1] => iter[1]
iter[1, 0, 1] => iter[1, 0]
```

It may look like this "doesn't test as much" as a more complex test case, but code coverage will tell you if you really hit all the cases.

conartist6 (changes requested):
And lastly you'll need to update `CHANGELOG`.

Overall it looks quite good! Most of my comments are just little things.

API.md
conartist6: It does look like the implementation is correct, it's just the example output that's wrong.

hillin: > distinct((x) => x % 7, [14, 21, 9, 777]); // Iterable[14, 9]

Nah, all of 14, 21 and 777 has a modulo 7 result of 0. Modulo gives you the remainder after dividing.

I'll try to find simpler yet still meaningful examples.

src/impls/$distinct/$distinct.js
hillin: Sure. The currying code is a little bit confusing but yet quite smart.

API.md
conartist6: Right. Meaning that of those numbers only 14 should appear in the output, because the other numbers are considered identical to it. 9 also makes it in because it's the first value for which the `x % 7 === 2`.

conartist6: The output can only contain values that were in the input (not 0 or 2).

src/impls/$distinct/$distinct.js
conartist6: I should probably add some better comments to it.

API.md
hillin: Oh, you are definitely right.

conartist6: I actually think a simpler version of that modulo example might be an excellent one to use since it demonstrates that property which you've just proven may seem surprising.

hillin: fixed in 45b72934c52c0df4e907d05d07b1429a8bc7dc43

src/impls/$distinct/__tests__/$distinct.test.ts
hillin: fixed in 45b72934c52c0df4e907d05d07b1429a8bc7dc43

src/impls/$distinct/__tests__/$distinct.test.ts
hillin: fixed in 45b72934c52c0df4e907d05d07b1429a8bc7dc43

src/impls/$distinct/$distinct.js
hillin: fixed in 64f50880549afd41939f95942a41006ffd65ce5d

src/impls/$distinct/DOCME.json
hillin: fixed in 64f50880549afd41939f95942a41006ffd65ce5d

hillin:
`CHANGELOG` updated in 64f50880549afd41939f95942a41006ffd65ce5d

API.md
```diff
@@ -437,6 +438,27 @@ append(4, [1, 2, 3]); // Iterable[1, 2, 3, 4]
 
 See [append](#append)
 
+### distinct
+
+**distinct(selector, [source](#wrappable))**  
+**distinct([source](#wrappable))**  
+**__distinct([source](#iterable), ?selector)**  
+
+Yields only the first occurrance of each value from `source`. An optional `selector` can be used to specify a transform function which returns a key to distinguish the values.
```
conartist6, line 447: Sorry to ask for this one last change, but I'm pretty picky about my docs. I think the docs should say this:
```
Defaults:
- `selector`: `value => value`

Yields values from `source` with duplicate values eliminated. A value is considered a duplicate if `Object.is(selector(a), selector(b))`. A `Set` is used to do this check in constant time.
```

conartist6 (approved):
I'm approving this because it is usable, but per the comment there's one more change I want before I merge.

API.md
conartist6: What do you think is the best way to handle stuff like this? I guess I could just edit your branch if you've granted me perms. Unfortunately it seems like as a maintainer it's difficult to tell if you've been granted perms to edit PRs. It seems you can find out by opening the online text editor and if you don't have perms it'll tell you that you'd have to make changes in a fork. That's kind of dumb, but OK. It looks like you've given me those perms. Is it good etiquette to use them? Rather than just making changes you don't know about and merging them I'd rather be opening a PR to your PR, and then if you see my changes and agree with them we could merge both at the same time. I've never tried doing that though.

hillin: I'm totally okay if you push to my fork, I didn't do anything to grant you the permission so I guess it's Github's default policy.

> A `Set` is used to do this check in constant time

This is not accurate, the `has()` method of `Set` works in logarithmic time (`O(logn)`), not constant time (`O(1)`).

conartist6: What's your source for `logn` time?

hillin: Sorry, that's embarrassing, I have to take that one back. I do understand that typical (Hash)Set implementations have a time complexity of `O(1)`, but somehow I remember someone has made a benchmark and proved that it was actually `O(logn)`. But now I turn to look for that post I can't find it. Guess my memory betrayed me!

conartist6: I mean that's the thing with JS -- none of the perf is guaranteed. The spec doesn't say one way or the other. I don't actually have any evidence for `O(1)` either, but that's what I was taught about how hash maps should work. Anyway OK, let's update the docs for now and push this out so you can use it.

hillin: Yeah specs say it's sublinear but I guess it's not a good idea to use that word in the docs.

Are you gonna make the changes or should I?

conartist6: I'll do it tomorrow

hillin: ping

conartist6:
Accepted as 5f6758

conartist6:
I couldn't merge this PR itself because of the branch protection rules on your `trunk` -- I couldn't update that branch in your repo because it would rewrite history. That's of course something that your repo inherited from this repo, but that's generally exactly what I want so it was easier to just merge the commit manually. In the future I'll try to make sure that PRs are made from unprotected branches.

conartist6:
Released in `v7.4.0`

PR #486, opened 2022-09-12
draft: min, minBy, max, maxBy operators

alexd6631:


conartist6:
I think getting the name right is a big part of it. I like your attention to not creating confusion with Rust's names for things, but I think `select` as a name could create just as much confusion as that term is already wildly overloaded. What about `findBest`? I think that would make the similarities to `find` clear, make it easier to discover the method, and would avoid confusion with SQL or async I/O.

src/impls/$select-by/__tests__/$select-by.test.ts
```diff
@@ -0,0 +1,58 @@
...
+        expect($await($selectBy((i) => i, minSelector, $wrap([1])))).toEqual(1);
+      }),
+    );
+  });
+
+  describe('when source has values', () => {
+    it(
+      'should return the best value according mapper and selector function',
+      $async(() => {
+        expect($await($selectBy(personAge, maxSelector, $wrap(PERSONS)))).toEqual(PERSONS[2]);
```
conartist6: I'd make this even more explicit, so the test is as clear as possible. I think you should do:
```js
const youngest = {
  name: 'a',
  age: 10,
};
const middle = {
  name: 'b',
  age: 20,
};
const oldest = {
  name: 'c',
  age: 30,
};

const people = [youngest, middle, oldest];
```
Then this assertion can be `toEqual(youngest)` which will be quite clear.

src/impls/$select-by/$select-by.js
```diff
@@ -0,0 +1,30 @@
...
+      }
+      $await(peekr.advance());
+    }
+    return bestItem;
+  }
+}
+
+export const $selectBy = $iterableCurry($__selectBy, {
+  reduces: true,
+});
```
conartist6: If `mapper` is optional you'll need
```
{
  reduces: true,
  minArgs: 1,
  maxArgs: 2,
}
```

src/impls/$select-by/$select-by.js
```diff
@@ -0,0 +1,30 @@
+import { $async, $await } from '../../../generate/async.macro.cjs';
+
+import { $iterableCurry } from '../../internal/$iterable.js';
+import { $__peekerate } from '../$peekerate/$peekerate.js';
+
+$async;
+export function $__selectBy(iterable, selector, mapper) {
```
conartist6: I think you should make `mapper` optional, with a default of `value => value`.

src/impls/$select-by/$select-by.js
```diff
@@ -0,0 +1,30 @@
...
+  const peekr = $await($__peekerate(iterable));
+
+  if (!peekr.done) {
+    let bestItem = peekr.value;
+    let bestValue = mapper(bestItem);
```
conartist6:
```js
let bestValue = $await(mapper(bestItem));
```

src/impls/$select-by/$select-by.js
```diff
@@ -0,0 +1,30 @@
...
+
+    $await(peekr.advance());
+    while (!peekr.done) {
+      const candidate = peekr.value;
+      const candidateValue = mapper(candidate);
```
conartist6:
```js
const candidateValue = $await(mapper(candidate));
```

src/impls/$select-by/$select-by.js
```diff
@@ -0,0 +1,30 @@
...
+
+import { $iterableCurry } from '../../internal/$iterable.js';
+import { $__peekerate } from '../$peekerate/$peekerate.js';
+
+$async;
+export function $__selectBy(iterable, selector, mapper) {
+  const peekr = $await($__peekerate(iterable));
+
+  if (!peekr.done) {
+    let bestItem = peekr.value;
```
conartist6: I've explicitly avoided using the term "item" in my code and documentation. I believe `bestItem` should be called `value` and `bestValue` should be `mappedValue`.

conartist6 (commented):
I think you're on the right track; I anticipate being able to merge this if you follow the suggestions I've made. You have (mostly) done a great job of copying existing patterns, so thanks for that!

If you don't mind I'll supply the documentation, as I'm rather picky about my docs and I don't want to make you jump through hoops to write things exactly the way that I would.

Do you have any feedback for me about how the contribution process could be improved?

conartist6:
A tip for doing the name change: first run `yarn clean:generated` to remove all generated files. Then do the rename and run `yarn generate` to recreate the generated files.

alexd6631:
> I think getting the name right is a big part of it. I like your attention to not creating confusion with Rust's names for things, but I think `select` as a name could create just as much confusion as that term is already wildly overloaded. What about `findBest`? I think that would make the similarities to `find` clear, make it easier to discover the method, and would avoid confusion with SQL or async I/O.

Hello @conartist6, 

To be fair, the more I think about it, finding the best element in a collection given a criteria, is exactly the definition of finding a maximum of that collection, given an ordering/comparator function. 

So I think we should stick with the common naming convention of `max` & `maxBy` since it is familiar across a variety of languages and libraries. Here are a few examples  (Rust, Kotlin, Js libs, Haskell) where the naming is consistently used.

 - https://doc.rust-lang.org/stable/std/iter/trait.Iterator.html#method.max
 - https://doc.rust-lang.org/stable/std/iter/trait.Iterator.html#method.max_by
 - https://doc.rust-lang.org/stable/std/iter/trait.Iterator.html#method.max_by_key
 - https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/max.html
 - https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/max-by.html
 - https://lodash.com/docs/4.17.15#max
 - https://lodash.com/docs/4.17.15#maxBy
 - https://immutable-js.com/docs/v4.1.0/Collection/#max()
 - https://immutable-js.com/docs/v4.1.0/Collection/#maxBy()
 - https://hackage.haskell.org/package/base-4.17.0.0/docs/GHC-OldList.html#v:maximumBy

Besides, max/maxBy is also as generic/expressive as the proposed select/findBest because you can pass a custom ordering function.

I would propose the following signature and ergonomics : 

```typescript
- max<T>(ordering: (a: T, b: T) => boolean = (a, b) => a < b, source: Iterable<T>): T | undefined
- maxBy<T, K>(mapper: (T) => K, ordering: (a: K, b: K) => boolean = (a, b) => a < b, source: Iterable<T>): T | undefined
``` 

The default ordering could be the native `<` js operator, which works on strings and number (also it doesn't crash on objects, but the return value doesn't makes a lot of sense) 

```
execPipe([1, 2, 3], max()) // should yield 3
execPipe([1, 2, 3], max((a, b) => a > b)) // should yield 1
execPipe(persons, max((p1, p2) => p1.age < p2.age) // should yield the oldest
execPipe(persons, maxBy(p => p.age)) // should yield the oldest
execPipe(persons, maxBy(p => p.age, (a, b) => a > b)) // should yield the youngest
```

We could argue we could unify the two functions by taking maxBy and making the mapper argument optional, and equals to the identity function by default. However I am not sure it will be as convenient (notably in typescript) for inferring the type and autocompletion since the first lambda you will write will be ambiguous to the inference engine).

The implementation would be almost identical as the current PR content, the difference is that I find it more natural to express the ordering in terms of "less than" < operator, but is up to debate.

Also, while you can theoretically reimplement min/minBy by reversing the ordering function, IMHO I think we should just provide them out of the box, since it is very convenient and I would personally expect if from an iterator library.

Looking forward for your feedback

src/impls/$select-by/$select-by.js
alexd6631: If we do that, it would make it equivalent to the select function, so it would be redundant. As mentioned in the comment, while we could have a single function with both mapper/selector optional, I think it would not play well with type inference/completion in the editor.

src/impls/$select-by/$select-by.js
alexd6631: Does that mean the mapper function could return a promise in the async implementation ?
I guess I should also update the typings to reflect this ?

With a FP background, I am not too fond of having mapper functions being asynchronous, possibly side-effecting (I think these kind of functions should be pure 99% of time), but it would be more flexible to allow so.

conartist6:
We do already have `takeSorted`, so you could do:

```js
first(takeSorted((a, b) => a.age > b.age, persons));
```

We could add `takeSortedBy` which would have a selector so that it could be:

```js
first(takeSortedBy(p => p.age, persons));
```

We already supply as a default the comparator used by `Array.sort`, namely `(a, b) => (a > b ? 1 : b > a ? -1 : 0);`

conartist6:
Looks like the place where my docs reference the default comparator is broken. I'll fix that.

conartist6:
Oh but `takeSorted` is a little awkward because it uses a heap internally, so to avoid unnecessary memory usage my example would have to be `first(takeSortedBy(p => p.age, 1, persons));`.

How would you feel about `takeMin`, and `takeMax` as names?

src/impls/$select-by/$select-by.js
conartist6: Yes, it does. I have chosen to allow that throughout the library. I expect that most users will not need or want to use that feature, but it's there.

alexd6631:
I was writing that and I saw your follow up message : 
```
While the first/takeSorted would be equivalent, I don't think it as efficient as the proposed operator. Since inserting into the heap is O(log n) and you are doing it n times, I think the complexity would be O(n * log(n)) vs the linear approach. 
``` 
and as you mentioned there is an additional memory requirement.

alexd6631:
> How would you feel about takeMin, and takeMax as names?

To the risk of looking stubborn (and I definitely am BTW :)), I don't see any particular reason not to stick with common established convention of max/maxBy.

To give a (bit caricatural) example it would be like naming the `map` operator `transform`, while it would make some sense, it would be confusing not to use the common convention for this operator, especially for people coming from other environment.

But you are the author here, and you have put a lot of work in it, so I think it is up to you to make the choice.

src/impls/$select-by/$select-by.js
conartist6: Hmm, idunno. I think I'd have to play with the real type definitions and see if there was an acceptable compromise. My typedefs do contain quite a few compromises, especially around currying, due to limitations of Typescript. I'm generally fine with this as iter-tools is a JS library first and a TS library second.

I'm hoping that for iter-tools 8 I'll be able to write a type for my currying function and drop all the hacked up custom overloads I've written. TS has relatively recently gained the power to express  such a currying function, but it's still gonna be a bear to write that type, and I've no guarantee the resultant error messages will make sense to human beings.

conartist6:
`takeSorted(1)` uses a min heap of size 1, which is essentially just a temporary variable. But you have to specify that `1` because the default is `Infinity` i.e. store all the value in the heap.

conartist6:
I'm quite stubborn too, but usually because I want to have the debate. I try to be open to changing my mind.

I think where I got into the `take*` naming convention was with `takeLast`. I chose this name to remind users that it would be necessary to traverse the entire iterable to get the last item, as well as to reserve the name `last` for a method which does not need to do that, which will only be possible to write if the language defines double-ended iterators. Otherwise I use take to denote a method like `takeSorted` that returns some slice or subset of values from the iterable.

Since neither of those situations applies to `min` and `max` though, I think those names are fine. In any world `min` needs to traverse its entire input, and the name should already make that obvious to any competent programmer

conartist6:
I think then that I'm giving you the green light for `min`, `max`, `minBy`, and `maxBy`, all with an optional comparator.

alexd6631:
> I think then that I'm giving you the green light for min, max, minBy, and maxBy, all with an optional comparator.

Great ! It is getting late here in France, so I will wrap the PR in the coming days and make your suggested changes.

alexd6631:
Hey 👋 

I renamed/implemented the max and maxBy operator by taking into account your remarks.
For the comparator function, I have gone with `(T, T) => number`, (instead of `(T, T) => boolean` previously) to be more consistent with takeSorted and other js libraries. This way I could reuse `defaultCompareOrder`.

If this looks good to you, I will implement min/minBy thereafter. One possible implementation strategy would be to reimplement min(By) by reusing max(By) and wrapping the user provided comparator in a comparator that invert its sign.

Do you prefer this approach, or should I copy the algorithm (it will be very similar only the comparison will be inverted and a few local variables renamed to make it clearer) ?

conartist6:
I'd prefer similar code.

alexd6631:
Hello,

I just pushed the implementation. Let me know if you have any more feedback, and how you want to proceed for the API doc.

conartist6:
Thanks! I'll look later today. I'm still planning to write the API docs.

conartist6:
Rust's docs say: "If several elements are equally maximum, the last element is returned."

Kotlin's says "Returns the first entry yielding the largest value of the given function" (and I double checked the implementation since that documentation text is only on a deprecated method).

Immutable: "If any values are comparatively equivalent, the first one found will be returned."

Lodash doesn't say what it does but it also returns the first value.

In ramda, [minBy](https://ramdajs.com/docs/#minBy) is not a collection method, and is meant to be used with reduce.

alexd6631:
I don't think there is a consensus on whether max should return the first or last maximum value encountered. As you have noticed in the implementation I only update the current candidate only if there is an improvement found, which will results in returning the first max value.

Personally I tend to think this an implementation detail, and max should only satisfy the invariant, that for all values in the iterator, they are less or equal than the returned maximum value, and I would not document / guarantee the order in the documentation. But if we are really sure we want to stick with this implementation we can document this behavior, as it can be helpful in some edge cases.

conartist6:
I have a few more thoughts from the past few days. At the moment I'm leaning towards going back to the select/findBest design.

First, consistency with existing methods which may or may not return a value dictates that you offer variants of your methods with an `or` suffix, the intent being to ensure that it is possible to code in a defensive style that is able to differentiate between finding the value `undefined` and not finding any value.

Fixing that means that we now go from 4 methods to 8 methods: `min`, `minOr`, `minBy`, `minByOr`, (by or? ew), `max`, etc. If we also differentiated between looking for the first-min versus last-min, we'd have 16 methods, and clearly things are starting to get absurd. When I've seen this name-explosion pattern occur in the past, it has always meant that the API I was offering was insufficiently flexible.

That's why I still really like the idea of a `findBest` API with a `valueIsNewBest` callback of `(best, value) => boolean`. I don't think you'd necessarily need or want a `findWorst` for that API. The point of the design is that the predicate allows any desired behavior to be specified. It can have an unambiguous optional `mapper` parameter, and the library could also export named helpers for the logical functions, so all in all usage would look like:

```js
import { findBest, lastHighest } from 'iter-tools-es';

const findBest(lastHighest, p => p.age, people);
```

I really like the usage because it you can understand everything important about what's happening at a glance.

Thoughts?

conartist6:
You had commented that the types for this API can be hard for typescript, e.g. there's no good way to type `findBest(predicate)(mapper)` because typescript needs the `mapper` type to know what inputs `predicate` would be receiving. I guess I'm not too worried about this. Typescript will be able to handle `findBest(predicate, mapper)` correctly, and that's the currying form I think people will actually be using.

conartist6:
I'm giving you write access so that we can collaborate on a branch. I want to see my idea mocked up, but I don't want to make you do all the work when I keep changing my mind about what the code should say.

alexd6631:
> First, consistency with existing methods which may or may not return a value dictates that you offer variants of your methods with an or suffix, the intent being to ensure that it is possible to code in a defensive style that is able to differentiate between finding the value undefined and not finding any value.

Personally I think the nullish coalescing operator (`??`) make those variants not that useful in modern JS and TS (and most of times you can transpile to older platform). I don't think the order is that important, and in some edge cases where it is there would be some workaround (like reversing the data source, or tweaking the comparator to take account of the ordering)

> You had commented that the types for this API can be hard for typescript, e.g. there's no good way to type findBest(predicate)(mapper) because typescript needs the mapper type to know what inputs predicate would be receiving. I guess I'm not too worried about this. Typescript will be able to handle findBest(predicate, mapper) correctly, and that's the currying form I think people will actually be using.

My point is that it will work technically, but the ergonomics may not be that great regarding to autocompletion, since you would like to write the mapper first, so that the types of predicate arguments are inferred and you have relevant autocompletion in the predicate lambda. Also if one of the function is optional, the first argument will be ambiguous and autocompletion will not be straightforward. Having a separate max/maxBy avoided these kind of ergonomics issues, and used with execPipe, the completion behavior was intuitive.

While I understand your concern about carefully designing this, notably avoiding bloating the library with useless variants, I think it would be a mistake not to include min/max (at least on the API surface level) for familiarity/convenience with other libs. I am not against having a advanced `findBest` operator for handling edge case, but I would still expose min/max(by) (maybe implemented in terms of findBest) because most of times, this is simply what you need.

conartist6:
The `or` suffixes aren't about null coalescing. `findBest(max, [undefined])` and `findBest(max, [])` both return `undefined`. so you actually can't tell from the return value which thing happened. Many usages don't need to make this distinction, but it's part of the library's functionality to ensure that it is possible.

Autocompletion is a bit of a concern I guess? The first argument is not ambiguous though -- the first argument is always the comparator. Notice I've used the `growRight` flag to my currying engine to ensure that. As you say though, autocomplete won't know the types of those variables yet until you give it the mapper. I don't see that as a big enough concern to direct the whole design, especially since my hope is that the presence of mappers means that the comparator can almost always be one of the ones I provide like `firstLowest`.

conartist6:
As to parity with other libraries, It definitely is one design concern, but it is not an overriding one. Eventually I'll have migration guides from other common and popular libraries, and now the guide can even tell you that if you're coming from Rust its `max` translates directly to `findBest(lastHighest)`, whereas if you're coming from Immutable `max` would translate directly to `findBest(firstHighest)`.

Another point in favor of `findBest` is that while it *may* be harder to write, I think it's easier to read. I do not envision people writing this so often that it's important that it need only be three letters. The bigger issue is that it's ambiguous whether the name `max` means a collection operation or comparing two numbers. The way I write code, at the top of the file I'd always write `const { max } = Math;` as opposed to writing `Math.max` inline. Now if I see `max` in my code, which thing is it? Whereas if I see `findBest` I'll know immediately that I'm traversing an iterable. This also may mean that it's easier to autocomplete `findBest`, because it's less likely to be ambiguous as to what it means.

Lastly my real highest design principle is completeness. I want to build things in such a way that they don't need to be rebuilt or revisited or reworked. When I design something, I want to ship it and move on. That means I need to serve all the use cases I can think of, and serve them equally well if possible. I like `findBest` because I may need to educate users about its use, but I'm less likely to have to tell them that they shouldn't use it or that it can't do what they need.

alexd6631:
Ok I think you made your mind here, and I am closing this PR.

conartist6:
Yeah, I'm sorry it worked out like that for you. I really appreciate the work you did putting this PR together -- thanks! I hope it won't be your last.

I'm sorry I couldn't give you the relevant feedback from the outset. I've mostly had my head in other projects recently, so it just took me some time to get my head back into how I build things here and why.

Also if you don't mind, I copy pasted some of your code to build my PR so I'd like to rebase my PR onto the commits of yours that introduce the code I used. That way you get credit for your work, and as a contributor to this repo.

PR #487, opened 2022-09-22
findBest

conartist6:
Competing pull request to https://github.com/iter-tools/iter-tools/pull/486, based on the discussions there.


conartist6:
Something to note: `findBest` works quite a lot like a `findLast` method as well. For example if you wanted to find out what the value of a single key in an iterable of entries (such as you would pass to `Map`) you could trivially use `findBest` to answer the question in a situation where you did not need or want the complete data structure:
```js
findBest(([key]) => key === 'key', entries);
```

conartist6:
Perhaps I should also add `findLast`? I've so far avoided adding methods that would be better defined as operations against reverse iterators though.

PR #497, opened 2023-05-23
Fix asyncBuffer signature docs

conartist6:
Supercedes #495

### DIFF
--- a/API.md
+++ b/API.md
@@ -2049,7 +2049,7 @@ See [takeLastOr](#takelastor)
 
 ### asyncBuffer
 
-**asyncBuffer(source, n)**  
+**asyncBuffer(n, source)**  
 **__asyncBuffer(source, n)**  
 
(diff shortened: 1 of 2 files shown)


conartist6:
Unfortunately my current system doesn't give me a great way to release this fix. The NPM package links directly to https://github.com/iter-tools/iter-tools/blob/v7.5.0/API.md. I can't change the NPM readme without releasing a new version of  the package, which I shouldn't be doing for doc fixes!

conartist6:
I guess I could update the `v7.5.0` tag as a hack, but then it doesn't truly point to the commit that was used to build the released version, which is also bad.

conartist6:
One decent solution would be to publish two tags for each release, i.e. `v7.5.0` and `d7.5.0`. The `v` prefixed tags would denote commits used to build released versions, and the `d` prefixed tags would start off the same as the `v` ones but could be safely bumped in PRs like this one. The API docs link in README would be: https://github.com/iter-tools/iter-tools/blob/d7.5.0/API.md.

conartist6:
Oh i need to fix the badge too, so maybe I'll do a patch version that resolves those two things, even though the library code won't be changing at all. I guess I'm OK with that if I'm not likely to have to do it again regularly.

PR #500, opened 2023-05-23
dX.Y.Z tags to allow doc updates

conartist6:
Fixes the problem that we can't fix typos and such in the API docs without releasing a new version of the package. We want to preserve the `v` tags to denote the commits published on npm, but the `d` tags will allow the linked docs to be directed elsewhere so that old docs can be corrected while still allowing the API docs for `trunk` to be in a between-versions state.


conartist6:
Subsequently changed to `dX.Y` since patch versions should not contain additions or changes to the public contract, i.e. the docs should be the same for all sibling patch versions.

PR #509, opened 2025-06-02
Fix missing ESM __modules

### LINKED ISSUE #508: Importing an individual __method using ESM
This feels like a question that I should be able to answer without posting an issue, but after looking everywhere I am about to draw the conclusion that this is simply not possible. But I fail to understand why that would make any sense, especially given the fact that both individual method imports and __methods serve the same purpose: resource efficiency. So here I am, hoping anyone can shed some light on this issue.

What I am trying to do is pretty straightforward, and I was expecting to be able to do so as follows:

```js
import __asyncWindow from 'iter-tools-es/__methods/async-window'
```

But this doesn't work, because `__methods` are only exported for CommonJS(??)

I would be interested in understanding:
1) whether my conclusion that this is not possible is correct;
2) if yes, what's the rationale behind this limitation;
3) if no, how to import an individual __method from a ES module.

Thanks in advance for your time.

conartist6:
Fixes #508

### DIFF
--- a/package.es.json
+++ b/package.es.json
@@ -14,7 +14,8 @@
       "types": "./methods/*.d.ts"
     },
     "./__methods/*": {
-      "require": "./__methods/*.js"
+      "require": "./__methods/*.js",
+      "import": "./__methods/*.js"
     },
     "./explode.macro": "./explode.macro.js",
     "./package.json": "./package.json"


timmolendijk:
Wow that's fast.

Yeah, that looks about right to me. Was assuming there was a reason for the way it was configured, but happy to learn it was merely an oversight.

Releasing this would help me, thanks!

conartist6:
OK, I put up `7.5.4` with this fix and one other.

